PR #27, opened 2016-05-12
Encrypt path in statfs() calls

lxp:
Paths in statfs() calls were not encrypted resulting in an `Function not implemented` error, when the unencrypted path didn't exist in the underlying (encrypted) filesystem.

```
$ df plain/existingdir
df: ‘plain/existingdir’: Function not implemented
```


rfjakob:
Looks good. Bonus points if you add TestStatFs to https://github.com/rfjakob/gocryptfs/blob/master/integration_tests/main_test.go ;)

rfjakob:
Merged as a93bcabe9c1dea14bf29b94cbddf62c1bec4d315 with the pull request text to the commit message.

PR #40, opened 2016-09-28
You can now run fuse on travis

chmduquesne:
Hi,

You can now run fuse on travis. This pull request should prove it, once travis has done its job.

I still have not succeeded in running fuse-xfstests, even on this project, but I assume it is possible! I managed to build it (see the commented lines).

I was wondering if you could add fuse-xfstests to the travis build, and I would be even more interested in seeing the same tests on the go-fuse loopbackfs!

Cheers,
Christophe-Marie

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,3 +19,21 @@ script:
   - go build
   - ./build.bash
   - if ! go version | grep go1.3.3 ; then go test ./internal/... ; fi
+  - mkdir /tmp/cipherdir /tmp/mnt
+  - echo 'pass' | gocryptfs -init /tmp/cipherdir
+  - echo 'pass' | gocryptfs /tmp/cipherdir /tmp/mnt
+  - echo 'bar' > /tmp/mnt/foo
+  - fusermount -u /tmp/mnt
+  - ls /tmp/cipherdir
+
+sudo: required
+dist: trusty
+before_install:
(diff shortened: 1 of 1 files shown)


chmduquesne:
Ok, I managed to actually run this test suite!

rfjakob:
Awesome! Note that you can just run ./test.bash, this runs the full integration test suite, which also mounts and unmounts several filesystems

chmduquesne:
All right, then I am sticking to the builtin integration test suite!

rfjakob:
By the way, have the travis people announced this somewhere?

chmduquesne:
Not really. I got it from [this comment](https://github.com/travis-ci/travis-ci/issues/1100#issuecomment-160169121), but the linked blog post does not mention fuse.

chmduquesne:
There must be some travic specificities that don't make this integration with fuse perfect, though. Because this test suite definitely passes on my machine.

rfjakob:
Interesting failure. Maybe the limit on the older system they are running is below 255.

--- FAIL: TestLongNames (0.00s)
	matrix_test.go:531: n255s is not in directory listing

rfjakob:
Fixed by https://github.com/rfjakob/gocryptfs/commit/67a959eebfc299f762fdb69283a4f0d46f28a977 . FUSE tests now run on Travis, thanks!

PR #42, opened 2016-10-02
lint fixes

vgough:
This fixes most of the warnings from "golint ./...", which are primarily warnings related to Go naming conventions and comment standards.

This fixes warnings from the following categories:
"don't use ALL_CAPS in Go names; use CamelCase"
"exported function XXX should have comment or be unexported"
"exported var XXX should have its own declaration"
"comment on exported function XXX should be of the form "XXX ...""
"exported func XXX returns unexported type *yyy, which can be annoying to use"
"if block ends with a return statement, so drop this else and outdent its block"

One class of golint warnings remains after this PR, which is due to package naming:
"don't use an underscore in package name"

### DIFF
--- a/tests/test_helpers/helpers.go
+++ b/tests/test_helpers/helpers.go
@@ -21,7 +21,7 @@ const testParentDir = "/tmp/gocryptfs-test-parent"
 // GocryptfsBinary is the assumed path to the gocryptfs build.
 const GocryptfsBinary = "../../gocryptfs"
 
-// TmpDir is used by "go test" running package test s in parallel! We create a
+// TmpDir is a unique temporary directory. "go test" runs package tests in parallel. We create a
 // unique TmpDir in init() so the tests do not interfere.
 var TmpDir string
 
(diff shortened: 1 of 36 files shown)


tests/test_helpers/helpers.go
```diff
@@ -17,16 +17,18 @@ import (
...
 // TmpDir will be created inside this directory
 const testParentDir = "/tmp/gocryptfs-test-parent"
+
+// GocryptfsBinary is the assumed path to the gocryptfs build.
 const GocryptfsBinary = "../../gocryptfs"
 
-// "go test" runs package tests in parallel! We create a unique TmpDir in
-// init() so the tests do not interfere.
+// TmpDir is used by "go test" running package test s in parallel! We create a
+// unique TmpDir in init() so the tests do not interfere.
```
rfjakob: TmpDir is a unique temporary directory. "go test" runs package tests in parallel. We create a unique TmpDir in init() so the tests do not interfere.

vgough:
Applied changes from the comments above, please take a look.

rfjakob:
Merged as https://github.com/rfjakob/gocryptfs/commit/b764917cd5c1b1d61b8ce08e7af0b29793fbbb80 , thanks!

PR #88, opened 2017-03-07
Report correct symbolic link dentry sizes

m-v-b:
Hello @rfjakob,

As mentioned in the commit message, I found a bug in `gocryptfs` relating to the symbolic link directory entry sizes in reverse-mounted gocryptfs file systems, and I implemented a fix for it. The issue relates to the size of the symbolic link dentry not being equal to the length of the target path string of the symbolic link.

Please bear with me if there are any glaring issues with the code changes; I have never used go, at least until the time I tried to fix this bug.

Thank you!


rfjakob:
Looks good, I'll review it in detail later today. Bonus points for the fact that you added a test :)

rfjakob:
Merged as d48ccb3ddab71773a991b8f1b062901ff5b435b0 , thanks!

PR #94, opened 2017-03-25
Further explain the use of 'scryptn' parameter

danim7:
Just a small PR to further explain the use of 'scryptn' parameter by telling the possible values.

Actually, I was playing around with the parameters of gocryptfs, and I had to debug the code to understand that the command line parameter is expected to be the exponent instead of the final value.

The error I was getting was a bit confusing:

```
gocryptfs -init -scryptn 2048 ram_ciphertext/
Choose a password for protecting your files.
Password: 
Repeat: 
2017/03/25 14:30:58 DeriveKey failed: scrypt: N must be > 1 and a power of 2
panic: DeriveKey failed: scrypt: N must be > 1 and a power of 2
```

### DIFF
--- a/Documentation/MANPAGE.md
+++ b/Documentation/MANPAGE.md
@@ -200,9 +200,9 @@ Options:
 	For more details visit https://github.com/rfjakob/gocryptfs/issues/92 .
 
 **-scryptn int**
-:	scrypt cost parameter logN. Setting this to a lower value speeds up
-	mounting but makes the password susceptible to brute-force attacks
-	(default 16)
+:	scrypt cost parameter logN. Possible values: 10-62. Setting this to a lower
+        value speeds up	mounting and reduces its memory needs, but makes
+	the password susceptible to brute-force attacks	(default 16).
 
 **-speed**
 :	Run crypto speed test. Benchmark Go's built-in GCM against OpenSSL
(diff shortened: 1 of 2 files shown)


rfjakob:
Nice, thanks. It looks like the error messages and the valid range depends on the version of x/crypto/scrypt. I get this for 29:

```
$ gocryptfs -init -scryptn 29 a
Choose a password for protecting your files.
Password: 
Repeat: 
panic: runtime error: makeslice: len out of range

goroutine 1 [running]:
golang.org/x/crypto/scrypt.Key(0xc420015440, 0x4, 0x8, 0xc42000d160, 0x20, 0x20, 0x20000000, 0x8, 0x1, 0x20, ...)
	/home/jakob/go/src/golang.org/x/crypto/scrypt/scrypt.go:235 +0x14c
```

danim7:
Actually, I calculated the max value given the N int type of the ScryptKDF struct and the transformation done: s.N = 1 << uint32(logN) in internal/configfile/scrypt.go

The purpose is to make clear that we don't expect a -scryptn of 2048 but a -scryptn of 20.

rfjakob:
If the documentation says "62" people will try to use it ;)

I suggest to say "28" is the maximum. This is accepted by scrypt but fails with an out-of-memory error later.

PS: The actual limits come from lines 230 and 235 in https://github.com/golang/crypto/blob/master/scrypt/scrypt.go

danim7:
Sure, changed to 28 ;)

PR #96, opened 2017-03-26
This is working on macos

jekriske-lilly:
This implements some of the changes discussed in #15 
The LDFLAGS and CFLAGS will work on OSX and if the paths don't exist in Linux the lines are just ignored of course.

Another addition of specifying a volname for osxfuse. If I use `gocryptfs cypher plain` then the resulting volume should be named 'plain' just as it would be on Linux.

This works in macos 10.12.3, Ubuntu 16.10, and Centos 7.3

### DIFF
--- a/internal/stupidgcm/stupidgcm.go
+++ b/internal/stupidgcm/stupidgcm.go
@@ -4,6 +4,8 @@
 // decryption functions. It only support 32-byte keys and 16-bit IVs.
 package stupidgcm
 
+// #cgo LDFLAGS: -L/usr/local/opt/openssl/lib
+// #cgo CFLAGS: -I/usr/local/opt/openssl/include
 // #include <openssl/evp.h>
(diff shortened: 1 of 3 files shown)


rfjakob:
Hi Jeff, I have commited the volname change as https://github.com/rfjakob/gocryptfs/commit/9f8e19b8565ba49922aaddd3935b846feb6dc5a3 , thanks!

I'm not sure about the openssl thing. I think most MacOS users will use "./build-without-openssl.bash" anyway as there are no disadvantages.

For the load_osxfuse issue, to me it looks like this should be added to go-fuse. There even is a TODO [in the code](https://github.com/hanwen/go-fuse/blob/master/fuse/mount_darwin.go#L23): 

    // TODO(hanwen): run the load_osxfuse command.

I'd suggest to continue that discussion at https://github.com/rfjakob/gocryptfs/issues/124 .

Issue #95, opened 2017-03-25
Confusing file owner for longname files in reverse mode


danim7:
Not sure if this shall be called a bug, or it is just an obscure use-case, but it is somewhat confusing:

1. I have two folders (note that they are owned by root and have 0777 permissions):
```
yo@wtf3-virtual:/media/gocryptfs$ ls -l
total 8
drwxrwxrwx 2 root root 4096 mars  25 18:38 ciphertext
drwxrwxrwx 2 root root 4096 mars  25 18:38 plaintext
```
2. Mount it in reverse mode:
`gocryptfs -reverse  plaintext/ ciphertext/`

3. In the plaintext folder, we have files with short and long filenames
```
yo@wtf3-virtual:/media/gocryptfs$ ls -l plaintext/
total 8
-rw-rw-r-- 1 yo yo 3 mars  25 18:38 12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890
-rw-rw-r-- 1 yo yo 3 mars  25 18:37 aa
```

4. In the ciphertext folder, we have the expected files...however, the file owner is a bit confusing, because for the *.name and the gocryptfs.diriv files it doesn't match the owner in the plaintext folder:
```
yo@wtf3-virtual:/media/gocryptfs$ ls -l ciphertext/
total 20
-r-------- 1 yo   yo   413 mars  25 18:36 gocryptfs.conf
-r-------- 1 root root  16 mars  25 18:38 gocryptfs.diriv
-rw-rw-r-- 1 yo   yo    53 mars  25 18:38 gocryptfs.longname.1HKAT_9tvS0womyymfnI-nynl5tJU8YBdk3Q3kPX0IU
-r-------- 1 root root 278 mars  25 18:38 gocryptfs.longname.1HKAT_9tvS0womyymfnI-nynl5tJU8YBdk3Q3kPX0IU.name
-rw-rw-r-- 1 yo   yo    53 mars  25 18:37 nhcZwS0V6sK6--GztikDYQ
```
5. However, even if with the "yo" user it shouldn't be possible, I can successfully read the files owned by root:
```
yo@wtf3-virtual:/media/gocryptfs$ md5sum ciphertext/*
38077a320b8c776d9ebb889049aec647  ciphertext/gocryptfs.conf
c61bd575cfbb88f33838ab94299f828e  ciphertext/gocryptfs.diriv
[...]

PR #97, opened 2017-03-27
Consistent file owners for virtual files in reverse mode

danim7:
This PR addresses the Issue #95 , about "Confusing file owner for longname files in reverse mode".

It affects only the reverse mode, and introduces three main modifications:
1. The "gocryptfs.diriv" files are assigned the owner and group of the gocryptfs process. In my opinion, these are virtual files, and they shall be owned by the user who mounted the gocryptfs folder and is therefore interested in reading it.
2. The "gocryptfs.longname.XXXX.name" files are assigned the owner and group of the underlying plaintext file. Therefore it is consistent with the file "gocryptfs.longname.XXXX" that has the encrypted contents of the plaintext file.
3. The two virtual files mentioned above are given -r--r--r-- permissions. This is consistent with the behavior described in function Access in internal/fusefrontend_reverse/rfs.go where all virtual files are always readable. Behavior also observed in point c) in #95 .

Please don't hesitate to tell me what you think and how to improve it, it is my first time writing code in Go language ;)

### DIFF
--- a/internal/fusefrontend_reverse/reverse_longnames.go
+++ b/internal/fusefrontend_reverse/reverse_longnames.go
@@ -96,13 +96,6 @@ func (rfs *ReverseFS) newNameFile(relPath string) (nodefs.File, fuse.Status) {
 		return nil, fuse.ToStatus(err)
 	}
 	content := []byte(rfs.nameTransform.EncryptName(e, dirIV))
-	parentFile := filepath.Join(rfs.args.Cipherdir, pDir)
-	
-	fi, err := os.Stat(filepath.Join(rfs.args.Cipherdir, pDir, e))
-	if err != nil {
-		return nil, fuse.ToStatus(err)
-	}
-	st := fi.Sys().(*syscall.Stat_t)
-
-	return rfs.newVirtualFile(content, parentFile, st.Uid, st.Gid)
+	parentFile := filepath.Join(rfs.args.Cipherdir, pDir, e)
+	return rfs.newVirtualFile(content, parentFile)
 }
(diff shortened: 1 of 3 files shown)


rfjakob:
Thanks for the bug report (very nice and detailed) and also the pull request! I agree with your conclusions and will review the pull request today in the evening.

internal/fusefrontend_reverse/virtualfile.go
```diff
@@ -59,8 +73,10 @@ func (f *virtualFile) GetAttr(a *fuse.Attr) fuse.Status {
 	}
 	st.Ino = f.ino
 	st.Size = int64(len(f.content))
-	st.Mode = syscall.S_IFREG | 0400
+	st.Mode = virtualFileMode
 	st.Nlink = 1
 	a.FromStat(&st)
+	a.Owner.Uid = f.Uid
+	a.Owner.Gid = f.Gid
```
rfjakob: This is not needed. The owner should already be set correctly from "Lstat(f.parentFile, &st)"

rfjakob: The new Uid and Gid attributes of virtualFile are redundant, then

danim7: That is the case for the gocryptfs.diriv files, but I need those fields for the gocryptfs.longname.*.name files. How should I do if not ?

rfjakob: You get them from parentFile

internal/fusefrontend_reverse/reverse_longnames.go
```diff
@@ -97,5 +97,12 @@ func (rfs *ReverseFS) newNameFile(relPath string) (nodefs.File, fuse.Status) {
 	}
 	content := []byte(rfs.nameTransform.EncryptName(e, dirIV))
 	parentFile := filepath.Join(rfs.args.Cipherdir, pDir)
```
rfjakob: Replace this with 

    parentFile := filepath.Join(rfs.args.Cipherdir, pDir, e)

danim7: Oh, I got it, just give me a minute

Issue #53, opened 2016-10-22
additional options (UID, GID, UMASK)


tolotos-rgu:
How can I use this options on mount: -ko uid=$USER_ID,gid=$GROUP_ID,umask=$UMASK_PROP?
All files copied into plain will chown $USER_ID:$GROUP_ID and chmod using $UMASK_PROP in plain and cipher.

rfjakob:
Hmm, none of these options is understood by the kernel.

You have the uid thing by starting gocryptfs as the user you want the files to end up with:
``` 
sudo -u USER gocryptfs -allow_other a b
```
Umask is a client-side thing, the files will end up with the umask the process has that created it.

charles-dyfis-net:
@rfjakob, ...if I wanted to implement uid/gid options to allow gocryptfs with allow_other to be used in situations where the user to be permitted to access a given tree and the user with OS-level write permissions to the backing store differ, would such a patch have a prayer of a chance of being merged? (For context, see https://serverfault.com/questions/852324/configuring-nfsv4-anonymous-ownership-mapping-idmapd-conf-per-subtree-or-mount -- I have a single mount from a NFS server, for which all files appear to be owned by `nfsnobody`, and want to expose different subtrees to different applications with different ownership).

PR #114, opened 2017-05-30
Implement force_owner option to display ownership as a specific user.

charles-dyfis-net:
This is a subset of what was previously proposed in #53. It permits a gocryptfs store backed by storage owned by user A (and, potentially, with filesystem-level mount options coercing ownership to user A) to be presented as readable and writable to user B, should user A invoke gocryptfs with `force_owner` passing user B's UID and GID.

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -217,6 +217,11 @@ func initFuseFrontend(key []byte, args *argContainer, confFile *configfile.ConfF
 	if args.aessiv {
 		cryptoBackend = cryptocore.BackendAESSIV
 	}
+	// forceOwner implies allow_other, as documented.
+	// Set this early, so args.allow_other can be relied on below this point.
+	if args._forceOwner != nil {
+		args.allow_other = true
+	}
 	frontendArgs := fusefrontend.Args{
 		Cipherdir:      args.cipherdir,
 		Masterkey:      key,
@@ -229,6 +234,7 @@ func initFuseFrontend(key []byte, args *argContainer, confFile *configfile.ConfF
 		HKDF:           args.hkdf,
 		SerializeReads: args.serialize_reads,
 		ForceDecode:    args.forcedecode,
+		ForceOwner:     args._forceOwner,
 	}
 	// confFile is nil when "-zerokey" or "-masterkey" was used
 	if confFile != nil {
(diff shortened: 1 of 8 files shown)


charles-dyfis-net:
(Test failure against go 1.6.4 also present on master; see https://travis-ci.org/rfjakob/gocryptfs/jobs/237577329).

charles-dyfis-net:
Just realized that I'd TODO'd reverse-mode support. Will be amending that in shortly; apologies about the oversight.

rfjakob:
Looks good, I'll request a small change, will review in a minute.

mount.go
```diff
@@ -229,6 +230,26 @@ func initFuseFrontend(key []byte, args *argContainer, confFile *configfile.ConfF
...
+			tlog.Fatal.Printf("force_owner: Unable to parse UID %v as positive integer", ownerPieces[0])
+			os.Exit(exitcodes.Usage)
+		}
+		gidNum, err := strconv.ParseInt(ownerPieces[1], 0, 32)
+		if err != nil || gidNum < 0 {
+			tlog.Fatal.Printf("force_owner: Unable to parse GID %v as positive integer", ownerPieces[1])
+			os.Exit(exitcodes.Usage)
+		}
+		frontendArgs.ForceOwner = &fuse.Owner{Uid: uint32(uidNum), Gid: uint32(gidNum)}
+		args.allow_other = true
```
rfjakob: Parsing looks good but should be moved to main. The problem with doing it here is that we may abort with a parse error **after** the user has entered the password. I think we should do all validation **before** asking the user to do work (here: prompt for the password) because that work is wasted when we abort.

I suggest you add a helper variable to argContainer like I have done with `_ctlsockFd`, for example named `_ForceOwner`. Then you can parse the force_owner string in main and store the result in _ForceOwner.

charles-dyfis-net:
@rfjakob, does this suit? While I ended up squashing it in (and rebasing on master to pick up the test fix), the individual change made in response to review feedback is (for posterity) 82f505910c7b92294faa642639206f70156cbd0e.

rfjakob:
Yes it does, but I think you lost allow_other on the way ;)

charles-dyfis-net:
Quite right! Folding fixup fa611fb1a7ba92c7b46483a552526de639e73e34 in.

rfjakob:
Merged, thanks!

PR #161, opened 2017-11-21
main: Add '-devrandom' commandline option

slackner:
Allows to use /dev/random for generating the master key instead of the
default Go implementation. When the kernel random generator has been
properly initialized both are considered equally secure, however:

* Versions of Go prior to 1.9 just fall back to /dev/urandom if the
  getrandom() syscall is blocking (Go Bug #19274)

* Kernel versions prior to 3.17 do not support getrandom(), and there
  is no check if the random generator has been properly initialized
  before reading from /dev/urandom

This is especially useful for embedded hardware with low-entroy. Please
note that generation of the master key might block indefinitely if the
kernel cannot harvest enough entropy.

### DIFF
--- a/Documentation/MANPAGE.md
+++ b/Documentation/MANPAGE.md
@@ -51,6 +51,13 @@ be suitable.
 #### -d, -debug
 Enable debug output
 
+#### -devrandom
+Use /dev/random for generating the master key instead of the default Go
+implementation. This is especially useful on embedded systems with Go versions
+prior to 1.9, which fall back to weak random data when the getrandom syscall
+is blocking. Using this option can block indefinitely when the kernel cannot
+harvest enough entropy.
+
 #### -extpass string
 Use an external program (like ssh-askpass) for the password prompt.
 The program should return the password on stdout, a trailing newline is
--- a/internal/configfile/config_file.go
+++ b/internal/configfile/config_file.go
@@ -5,7 +5,9 @@ package configfile
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"log"
 
 	"github.com/rfjakob/gocryptfs/internal/contentenc"
 	"github.com/rfjakob/gocryptfs/internal/cryptocore"
@@ -47,10 +49,25 @@ type ConfFile struct {
 	filename string
 }
 
+// randBytesDevRandom gets "n" random bytes from /dev/random or panics
+func randBytesDevRandom(n int) []byte {
+	f, err := os.Open("/dev/random")
+	if err != nil {
+		log.Panic("Failed to open /dev/random: " + err.Error())
+	}
+	defer f.Close()
+	b := make([]byte, n)
+	_, err = io.ReadFull(f, b)
+	if err != nil {
+		log.Panic("Failed to read random bytes: " + err.Error())
+	}
+	return b
+}
+
 // CreateConfFile - create a new config with a random key encrypted with
 // "password" and write it to "filename".
 // Uses scrypt with cost parameter logN.
-func CreateConfFile(filename string, password string, plaintextNames bool, logN int, creator string, aessiv bool) error {
+func CreateConfFile(filename string, password string, plaintextNames bool, logN int, creator string, aessiv bool, devrandom bool) error {
 	var cf ConfFile
 	cf.filename = filename
 	cf.Creator = creator
@@ -72,7 +89,12 @@ func CreateConfFile(filename string, password string, plaintextNames bool, logN 
 	}
 
 	// Generate new random master key
-	key := cryptocore.RandBytes(cryptocore.KeyLen)
+	var key []byte
+	if devrandom {
+		key = randBytesDevRandom(cryptocore.KeyLen)
+	} else {
+		key = cryptocore.RandBytes(cryptocore.KeyLen)
+	}
 
 	// Encrypt it using the password
 	// This sets ScryptObject and EncryptedKey
--- a/internal/configfile/config_test.go
+++ b/internal/configfile/config_test.go
@@ -60,7 +60,7 @@ func TestLoadV2StrangeFeature(t *testing.T) {
 }
 
 func TestCreateConfDefault(t *testing.T) {
-	err := CreateConfFile("config_test/tmp.conf", "test", false, 10, "test", false)
+	err := CreateConfFile("config_test/tmp.conf", "test", false, 10, "test", false, false)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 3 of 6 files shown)


slackner:
The command line option ('-use-random') was chosen similar to the equivalent option of cryptsetup ('--use-random').

rfjakob:
Maybe here: https://github.com/rfjakob/gocryptfs/blob/master/internal/configfile/config_file.go#L74
Read 1 byte from /dev/random, which will block until the RNG is ready. Note that this is not my idea, I read it in an article explaining urandom as one use case where using /dev/random make sense: blocking on it.

slackner:
I'm not sure if reading 1 byte from /dev/random is a sufficient check if the RNG is ready. It feels more like a hacky workaround.

Maybe an alternative would be to use the Linux syscall directly (basically the code after the Go 1.9 changes)? For older kernels you would still need a /dev/random based method though.

rfjakob:
Another problem with the 1 byte read is that it depletes the entropy pool by 1 byte, which may become a problem if you create lots of filesystems, like when running automated tests.

Ok let's add this flag. I'll review the code and add comments.

Documentation/MANPAGE.md
```diff
@@ -280,6 +280,13 @@ Run crypto speed test. Benchmark Go's built-in GCM against OpenSSL
 #### -trace string
 Write execution trace to file. View the trace using "go tool trace FILE".
 
+#### -use-random
+Use /dev/random for generating the master key instead of the default Go
+implementation. This is especially useful on embedded systems with Go versions
+prior to 1.9, which fall back to weak random data when the getrandom syscall
+is blocking. Using this option can block indefinitely when the kernel cannot
+harvest enough entropy.
+
```
rfjakob: The "use-random" flag would be the only one that uses a hyphen "-" instead of an underscore "_". While I understand that this is what cryptsetup does, I think it is more important to be self-consistent here. I personally would just call it
`-devrandom` because it's clear, short, and works without a separator.

internal/cryptocore/nonce.go
```diff
@@ -16,6 +18,18 @@ func RandBytes(n int) []byte {
...
+	f, err := os.Open("/dev/random")
+	defer f.Close()
+	b := make([]byte, n)
+	_, err = io.ReadFull(f, b)
+	if err != nil {
+		log.Panic("Failed to read random bytes: " + err.Error())
+	}
+	return b
+}
+
```
rfjakob: Because this function should not be used anywhere but in CreateConfFile, I think it should move to config_file.go and be unexported (lowercase first letter).

slackner:
Both issues should be fixed. I've updated the commit, please let me know if you have any other remarks. (And also if you prefer to have updates as separate commits. I've merged them into the first patch this time.)

rfjakob:
Whoops, seems like something still missing. I wonder why Travis did not report this.
```
0 jakob@brikett:~/go/src/github.com/rfjakob/gocryptfs$ ./test.bash 
gocryptfs v1.4.2-5-g44ddb35 without_openssl; go-fuse v20170619-24-g3d30ad6; 2017-11-21 go1.9.2
gocryptfs v1.4.2-5-g44ddb35; go-fuse v20170619-24-g3d30ad6; 2017-11-21 go1.9.2
# github.com/rfjakob/gocryptfs/internal/configfile
internal/configfile/config_test.go:63:23: not enough arguments in call to CreateConfFile
	have (string, string, bool, number, string, bool)
	want (string, string, bool, int, string, bool, bool)
internal/configfile/config_test.go:84:23: not enough arguments in call to CreateConfFile
	have (string, string, bool, number, string, bool)
	want (string, string, bool, int, string, bool, bool)
internal/configfile/config_test.go:105:23: not enough arguments in call to CreateConfFile
	have (string, string, bool, number, string, bool)
	want (string, string, bool, int, string, bool, bool)
ok  	github.com/rfjakob/gocryptfs	0.005s
?   	github.com/rfjakob/gocryptfs/gocryptfs-xray	[no test files]
FAIL	github.com/rfjakob/gocryptfs/internal/configfile [build failed]
ok  	github.com/rfjakob/gocryptfs/internal/contentenc	0.003s
ok  	github.com/rfjakob/gocryptfs/internal/cryptocore	0.207s
ok  	github.com/rfjakob/gocryptfs/internal/ctlsock	0.002s
?   	github.com/rfjakob/gocryptfs/internal/exitcodes	[no test files]
?   	github.com/rfjakob/gocryptfs/internal/fusefrontend	[no test files]
?   	github.com/rfjakob/gocryptfs/internal/fusefrontend_reverse	[no test files]
ok  	github.com/rfjakob/gocryptfs/internal/nametransform	0.013s
?   	github.com/rfjakob/gocryptfs/internal/nametransform/dirivcache	[no test files]
?   	github.com/rfjakob/gocryptfs/internal/openfiletable	[no test files]
ok  	github.com/rfjakob/gocryptfs/internal/pathiv	0.002s
ok  	github.com/rfjakob/gocryptfs/internal/prefer_openssl	0.003s
ok  	github.com/rfjakob/gocryptfs/internal/readpassword	0.027s
?   	github.com/rfjakob/gocryptfs/internal/serialize_reads	[no test files]
ok  	github.com/rfjakob/gocryptfs/internal/siv_aead	0.002s
ok  	github.com/rfjakob/gocryptfs/internal/speed	0.002s [no tests to run]
ok  	github.com/rfjakob/gocryptfs/internal/stupidgcm	1.660s
ok  	github.com/rfjakob/gocryptfs/internal/syscallcompat	0.010s
?   	github.com/rfjakob/gocryptfs/internal/tlog	[no test files]
ok  	github.com/rfjakob/gocryptfs/tests/cli	0.937s
ok  	github.com/rfjakob/gocryptfs/tests/defaults	0.688s
ok  	github.com/rfjakob/gocryptfs/tests/example_filesystems	3.441s
ok  	github.com/rfjakob/gocryptfs/tests/hkdf_sanity	0.108s
ok  	github.com/rfjakob/gocryptfs/tests/matrix	11.372s
ok  	github.com/rfjakob/gocryptfs/tests/plaintextnames	0.091s
ok  	github.com/rfjakob/gocryptfs/tests/reverse	0.734s
?   	github.com/rfjakob/gocryptfs/tests/test_helpers	[no test files]
2 jakob@brikett:~/go/src/github.com/rfjakob/gocryptfs$ 
```

slackner:
Whoops, sorry. Should be fixed now.

internal/configfile/config_file.go
```diff
@@ -47,10 +49,22 @@ type ConfFile struct {
 	filename string
 }
 
+// randBytesDevRandom gets "n" random bytes from /dev/random or panics
+func randBytesDevRandom(n int) []byte {
+	f, err := os.Open("/dev/random")
```
rfjakob: Please check the error from Open()

internal/configfile/config_test.go
```diff
@@ -60,7 +60,7 @@ func TestLoadV2StrangeFeature(t *testing.T) {
 }
 
 func TestCreateConfDefault(t *testing.T) {
-	err := CreateConfFile("config_test/tmp.conf", "test", false, 10, "test", false)
+	err := CreateConfFile("config_test/tmp.conf", "test", false, 10, "test", false, false)
 	if err != nil {
 		t.Fatal(err)
 	}
```
rfjakob: Please add a simple test case for the new parameter, this should be good enogh:

```
// Passing "-devrandom" enables the last argument. Test that it "just works"
// (does not hang).
func TestCreateConfDevrandom(t *testing.T) {
	err := CreateConfFile("config_test/tmp.conf", "test", false, 10, "test", false, true)
	if err != nil {
		t.Fatal(err)
	}
}
```

slackner:
Should be fixed. I've also added another test for the CLI option. Any other remarks?

rfjakob:
Looks good to me. Merging.

rfjakob:
Thanks!

slackner:
No problem, thanks for reviewing!

Issue #163, opened 2017-11-22
Insufficient path validation in reverse mode allows access to files outside of mount point


slackner:
Assume user A has a reverse GoCryptFS mount with `-allow_other` and user B has access to the same master key (but no access to other files of user A). User B can now compute `rfs.nameTransform.EncryptName("..", dirIV)` with the directory IV of the root dir and look at files outside of the mount point.

**Steps to reproduce:**
* Create a regular reverse mount point
* Manually encrypt ".." with the root directory IV
* Type `ls <mount>/<encrypted "..">` to get a list of files (encrypted) outside of the mount

In order to fix this issue `rDecryptName` should validate the decrypted path component and reject such exploit attempts. My suggestion would be to reject ".", ".." and all strings containing "/", or is there anything else?

slackner:
I believe this issue is fully fixed with c547673529cb4934ab885081f5682e85aa994f79, so I'm going to close this bug report. Feel free to reopen if you think there are more corner cases which have to be handled. I've opened another pull request for some additional tests.

PR #164, opened 2017-11-22
nametransform: Return error if decrypted name is '.' or '..'

slackner:
Proposed fix for issue #163.

### DIFF
--- a/internal/nametransform/names.go
+++ b/internal/nametransform/names.go
@@ -65,6 +65,10 @@ func (n *NameTransform) DecryptName(cipherName string, iv []byte) (string, error
 	if bytes.Contains(bin, []byte{0}) || bytes.Contains(bin, []byte("/")) {
 		return "", syscall.EBADMSG
 	}
+	// The name should never be "." or "..".
+	if bytes.Equal(bin, []byte(".")) || bytes.Equal(bin, []byte("..")) {
+		return "", syscall.EBADMSG
+	}
 	plain := string(bin)
 	return plain, err
 }


rfjakob:
I would merge this right away if Travis testing was working. I hope to get more info out of https://github.com/travis-ci/travis-ci/issues/8790

rfjakob:
I got a reply and this seems to be a false positive in their abuse detection system. The blocked requests are shown in a hidden tab: https://travis-ci.org/rfjakob/gocryptfs/requests .

As a workaround they suggest that you sign into https://travis-ci.org/ once (see below). Is this acceptable for you?

> I've been looking into these rejected requests and I see that the fds-team organization nor any of its members have been never logged into Travis CI and this is why we are currently blocking these requests that come from fds-team's fork of your rfjakob/gocryptfs repository. Sorry for the inconvenience.
> 
> We are looking into fixing this false positive in our abuse detection system, as we don't feel that an organization or account should have to create a Travis account when working with another project that is using Travis CI.
> 
> To sort this out in the meantime, do you know who manages this organization and do you think one of their members could try to log into Travis CI at least once at https://travis-ci.org/ and try closing and reopen one of the rejected Pull Requests?

slackner:
> As a workaround they suggest that you sign into https://travis-ci.org/ once (see below). Is this acceptable for you?

I just did that, hopefully it will work now. Nevertheless, I find it a bit weird that all contributors need Travis accounts?! I would still consider that a bug in their system.

rfjakob:
I have never seen that problem, maybe only happens when the repo is owned by an organization? Anyway, seems to work now, thanks!

Issue #168, opened 2017-11-23
Longname cache in reverse mode should use both dirIV and longname as key


slackner:
**Steps to reproduce:**

1. Create a regular reverse mount point
2. Create files with the same very long name in multiple directories - so far so good, it will appear with a different name each time, for example, `gocryptfs.longname.A` in directory A and `gocryptfs.longname.B` in directory B
3. Try to access a path with `A/gocryptfs.longname.B` or `B/gocryptfs.longname.A` - this should fail, but it actually works.

The problem is that the longname cache only uses the path as key, and not the dirIV. Assume an attacker knows the relation longname path -> unencoded path in one directory, it allows to test if the same unencoded filename appears in any other directory.

Another problem of this vulnerability is that it is possible to get multiple ciphertexts for the same plaintext. For example, both `A/gocryptfs.longname.A` and `A/gocryptfs.longname.B` are two different ciphertext representations of the same plaintext.

PR #169, opened 2017-11-24
fusefrontend_reverse: Do not mix up cache information for different directories

slackner:
Fixes https://github.com/rfjakob/gocryptfs/issues/168

I decided to use the dir instead of the dirIV because it seems a bit easier here (especially when the dirIV could be variable length).

### DIFF
--- a/internal/fusefrontend_reverse/reverse_longnames.go
+++ b/internal/fusefrontend_reverse/reverse_longnames.go
@@ -51,7 +51,7 @@ func initLongnameCache() {
 // findLongnameParent converts "gocryptfs.longname.XYZ" to the plaintext name
 func (rfs *ReverseFS) findLongnameParent(dir string, dirIV []byte, longname string) (plaintextName string, err error) {
 	longnameCacheLock.Lock()
-	hit := longnameParentCache[longname]
+	hit := longnameParentCache[dir + "/" + longname]
 	longnameCacheLock.Unlock()
 	if hit != "" {
 		return hit, nil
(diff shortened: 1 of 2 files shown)


slackner:
```
--- FAIL: TestExampleFSv07PlaintextNames (0.10s)
	helpers.go:258: directory not empty
	helpers.go:233: mkdir /tmp/gocryptfs-test-parent/298521282/v0.7-plaintextnames.mnt.2/dir1: file exists
```

Hm, I believe this is a random issue (also only affects one version). Closing & reopening to give it another try.

internal/fusefrontend_reverse/reverse_longnames.go
```diff
@@ -51,7 +51,7 @@ func initLongnameCache() {
 // findLongnameParent converts "gocryptfs.longname.XYZ" to the plaintext name
 func (rfs *ReverseFS) findLongnameParent(dir string, dirIV []byte, longname string) (plaintextName string, err error) {
 	longnameCacheLock.Lock()
-	hit := longnameParentCache[longname]
+	hit := longnameParentCache[dir + "/" + longname]
```
slackner: If you prefer a different solution here, feel free to let me know. I am aware that `dir` is an unencrypted path component, whereas `longname` is an encrypted path component. I still decided to use a concatenated path to avoid a separate struct. The only requirement is that the key is unique for all (dir, longname) pairs.

rfjakob:
Looks good to me, using "dir" is fine IMO.

Can you add the (excellent) description from https://github.com/rfjakob/gocryptfs/issues/168 to the commit message? This way one can understandy why this was done without needing access to the github ticket.

slackner:
Thanks! I've updated the commit message.

Issue #171, opened 2017-11-24
Rmdir fails when the directory does not contain a gocryptfs.diriv file


slackner:
**Steps to reproduce:**

1. Create a regular GoCryptFS mount (no reverse mode in this case!)
2. Create a new directory in the mount point
3. Manually delete the gocryptfs.diriv file from the corresponding ciphertext directory
4. Attempt to delete the directory `rmdir <dirname>`

It will fail with:
```
rmdir: failed to remove '<dirname>': No such file or directory
```

However, the directory was still deleted. Problem seems to be a missing `return` in the handling of empty directories:

https://github.com/rfjakob/gocryptfs/blob/master/internal/fusefrontend/fs_dir.go#L193

It probably should always return, not just in case of a failure.

slackner:
@rfjakob I've created pull requests for some of the issues I have found. I hope you aren't too annoyed by all the sudden activity on here. ;) Note that I also don't mind if you or someone else wants to help out with these bugs. When I open a bug report it means that I do not have a patch yet (at the time of writing), so other people are also invited to contribute ofc.

rfjakob:
Not annoyed at all, it's awesome. Even if there is no immediate solution, documenting the problem makes sense.

PR #173, opened 2017-11-25
fusefrontend: Skip gocryptfs.diriv handling when directory was deleted successfully

slackner:
Fixes https://github.com/rfjakob/gocryptfs/issues/171


slackner:
I've also updated the commit message here to include more details about the problem.

Issue #177, opened 2017-11-26
Race-conditions related to PreserveOwner allow to set owner/group of arbitrary files when using plaintextnames


slackner:
**Steps to reproduce:**

1. Create a forward mount point with `plaintextnames` enabled
2. Mount as root user with `allow_other`
3. For testing purposes create a file `/tmp/file_owned_by_root` which is owned by the root user
4. As a regular user run inside of the GoCryptFS mount:

```
mkdir tempdir
mknod tempdir/file_owned_by_root p &
mv tempdir tempdir2
ln -s /tmp tempdir
```

When the steps are done fast enough and in the right order (run in a loop!), the device file will be created in `tempdir`, but the `lchown` will be executed by following the symlink. As a result, the ownership of the file located at `/tmp/file_owned_by_root` will be changed.

slackner:
Note that this issue could be fixed by always opening a dirfd, and then using `fchownat`.

rfjakob:
Problem on MacOS:
```
+ GOOS=darwin
+ GOARCH=amd64
+ go build -tags without_openssl
# github.com/rfjakob/gocryptfs/internal/fusefrontend
internal/fusefrontend/fs.go:311:3: undefined: syscall.Fchownat
```
Needs a compat wrapper like https://github.com/rfjakob/gocryptfs/blob/master/internal/syscallcompat/sys_darwin.go#L92

slackner:
Regarding the compat wrappers in general: wouldn't it be better to use open(".") and store a FD instead of the path (at least where it is possible)? This also reduces the risk of race conditions.

rfjakob:
On macos, you mean?

slackner:
> On macos, you mean?

Yes. Instead of using `Getwd()` to store the previous working directory, you could just open the current directory with `open(".")` and later use `Fchdir`. This would also avoid race-conditions when directories are renamed in the meantime.

rfjakob:
Yes, good idea. Looks like the whole file can be simplified by moving the locking and chdir-back-logic into dirfdAbs, at a small performance cost for Renameat (one additional chdir).

slackner:
Not sure if that is a good idea. Besides rename, all other functions only expect one path argument. Instead of the conversion to an absolute path, you could just Chdir and then run the syscall with relative path. This is already done for `Openat` and probably faster than calling `Getwd`. The common logic could still be moved to a helper function of course.

rfjakob:
[...]

Issue #176, opened 2017-11-25
Symlink owner is not set in forward mode with plaintextnames enabled


slackner:
See the corresponding `Symlink()` implementation in `internal/fusefrontend/fs.go`. In all other code paths the owner is set, only here it is missing.

slackner:
Reference: https://github.com/rfjakob/gocryptfs/blob/master/internal/fusefrontend/fs.go#L428

Issue #174, opened 2017-11-25
Missing check for plaintextnames in many FUSE callbacks


slackner:
**Steps to reproduce:**

1. Create a regular GoCryptFS mount point in forward mode and with `-plaintextnames`
2. In the mount point directory try to run `touch gocryptfs.longname.XXX`

This fails with the error `No such file or directory`, but should actually succeed. The problem is that many FUSE calls do not explicitly check if `plaintextnames` is set.

slackner:
On the first sight this affects the following functions:

* [ ] Create
* [ ] Mknod
* [ ] Unlink
* [ ] Rename
* [ ] Link

slackner:
I fear this one was closed a bit too early. Mknod and Unlink are fixed now, but there might be three other functions affected by the same problem. Should I open new bugs for them?

rfjakob:
Was auto-closed through the commit message i guess, just re-open it

slackner:
I would, but apparently I don't have permission to reopen myself.

rfjakob:
Hmm will check if i can allow that somehow

Issue #165, opened 2017-11-22
Race condition when reading symlinks can expose unrelated parts of the file system in reverse mode


slackner:
Assume user A has access to a machine, and root has mounted an encrypted version of the user home directory with `-allow_other`. Further assume that the user knows the master key. By quickly exchanging directories with symlinks in the original data, it is possible trick GoCryptFS into exposing unwanted parts of the file system.

**Steps to reproduce:**

1. Create a regular reverse mount point
2. Create a directory "test" in the original directory
3. Access the corresponding encrypted directory in the mount point (`ls <encrypted dir>`)
4. Quickly delete the directory in the original data, and instead create a symlink pointing somewhere else, for example to `/root`
5. Access the encrypted directory again, it will now blindly follow the symlink and reveal the content of the target directory

I used a small script to reproduce this issue, manual execution is probably too slow. Even without `-allow_other` this could be a security issue and could, for example, backup unwanted data instead of just the symbolic link.

slackner:
I have not prepared a patch for this yet, but apparently there is a `O_NOFOLLOW` constant which could be used. This requires to handle symlinks in the cipherDir manually, before calling `open()`.

rfjakob:
I wonder how apache or nginx handle this

rfjakob:
I believe we can always use O_NOFOLLOW. The kernel resolves symlinks before we geht the path. In other words, we should never get a "benign" open on a symlink.

slackner:
> I believe we can always use O_NOFOLLOW.

I'm not sure if it is currently allowed, but please keep in mind that the cipherDir could also contain symlinks. We probably want to resolve them immediately after program startup if this is not done yet.

rfjakob:
Not sure what you mean. Symlinks look like random crap in reverse mode, and because the kernel resolves resolves them, you can never follow them:

```
$ gocryptfs -reverse a b

$ tree
.
├── a
│   ├── bar -> foo
│   └── foo
└── b
    ├── gocryptfs.conf
    ├── gocryptfs.diriv
    ├── HUcUNp8dgFr_3JyES4rEVg
    │   └── gocryptfs.diriv
    └── NUq2Y99YTs1eebDg8H3jrA -> hv8qARpFAnWa5AzBuy9sIOvlU2ZVapmbZ5iVxdvX7HZW3G8


$ cd b/NUq2Y99YTs1eebDg8H3jrA 
[...]

PR #178, opened 2017-11-28
Multiple fixes for PlaintextNames mode

slackner:
This set fixes:

* Some bugs in the recently introduced `Fchownat` syscall wrapper. Unfortunately I do not have any macOS machine available, so please verify that everything still works after these changes.

* Partially issue #177 (the symlink part)

* Issue #176

* Partially issue #174 (the mknod and unlink part)

In addition the set introduces a `openBackingPath` helper to do path conversion and open a handle to the directory. This could be especially useful later for fixing issue #165.

Please let me know what you think. Also feel free to cherry-pick certain patches if you want to solve certain issues in a different way.

### DIFF
--- a/internal/fusefrontend/fs.go
+++ b/internal/fusefrontend/fs.go
@@ -448,8 +448,7 @@ func (fs *FS) Symlink(target string, linkName string, context *fuse.Context) (co
 			return fuse.ToStatus(err)
 		}
 		// Create "gocryptfs.longfile." symlink
-		// TODO use syscall.Symlinkat once it is available in Go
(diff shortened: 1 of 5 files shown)


rfjakob:
Looks good to me, thanks!

rfjakob:
PS: About the MacOS syscall wrappers: the plan is to make them testable under Linux. Testing on MacOS is too much hassle.

slackner:
Thanks for reviewing! Offering a way to test the MacOS code path on Linux definitely sounds like a good idea. Those syscall wrappers are already pretty complicated, would be nice to integrate them into the automated testing framework.

PR #179, opened 2017-11-29
Use *at() syscalls for Rmdir, Chmod, Chown and Mkdir

slackner:
Patch 1 fixes an existing TODO in the code.

Patch 2 and 3 changes the implementation of Chmod and Chown, such that `openBackingPath` and *at() syscalls are used. The general idea would be that in the long term, we try to use `openBackingPath` everywhere. When we later want to harden gocryptfs against symlink race-conditions (maybe optional, since it probably will be a bit slower), it is sufficient to change only a single function, instead of adding checks everywhere in the code.

Patch 4 should fix the remaining issue from #177.

### DIFF
--- a/internal/fusefrontend/fs_dir.go
+++ b/internal/fusefrontend/fs_dir.go
@@ -11,6 +11,8 @@ import (
 	"sync"
 	"syscall"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/hanwen/go-fuse/fuse"
 
 	"github.com/rfjakob/gocryptfs/internal/configfile"
@@ -231,9 +233,7 @@ retry:
 		return fuse.ToStatus(err)
 	}
 	// Actual Rmdir
-	// TODO Use syscall.Unlinkat with the AT_REMOVEDIR flag once it is available
-	// in Go
-	err = syscall.Rmdir(cPath)
+	err = syscallcompat.Unlinkat(int(parentDirFd.Fd()), cName, unix.AT_REMOVEDIR)
 	if err != nil {
 		// This can happen if another file in the directory was created in the
 		// meantime, undo the rename
@@ -245,7 +245,7 @@ retry:
 		return fuse.ToStatus(err)
 	}
 	// Delete "gocryptfs.diriv.rmdir.XYZ"
-	err = syscallcompat.Unlinkat(int(parentDirFd.Fd()), tmpName)
+	err = syscallcompat.Unlinkat(int(parentDirFd.Fd()), tmpName, 0)
 	if err != nil {
 		tlog.Warn.Printf("Rmdir: Could not clean up %s: %v", tmpName, err)
 	}
--- a/internal/fusefrontend/fs_dir.go
+++ b/internal/fusefrontend/fs_dir.go
@@ -24,7 +24,7 @@ import (
 
 const dsStoreName = ".DS_Store"
 
-func (fs *FS) mkdirWithIv(cPath string, mode uint32) error {
+func (fs *FS) mkdirWithIv(dirfd *os.File, cName string, mode uint32) error {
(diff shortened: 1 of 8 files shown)


internal/syscallcompat/sys_linux.go
```diff
@@ -54,9 +54,19 @@ func Renameat(olddirfd int, oldpath string, newdirfd int, newpath string) (err e
 	return syscall.Renameat(olddirfd, oldpath, newdirfd, newpath)
 }
 
-// Unlinkat wraps the Unlinkat syscall.
-func Unlinkat(dirfd int, path string) error {
-	return syscall.Unlinkat(dirfd, path)
+// Unlinkat syscall. In old versions the 'flags' argument was missing, so
+// manually call it by using the corresponding syscall number.
+func Unlinkat(dirfd int, path string, flags int) (err error) {
```
rfjakob: Should get a test

internal/fusefrontend/fs_dir.go
```diff
@@ -24,22 +24,22 @@ import (
 
 const dsStoreName = ".DS_Store"
 
-func (fs *FS) mkdirWithIv(cPath string, mode uint32) error {
+func (fs *FS) mkdirWithIv(dirfd *os.File, cName string, mode uint32) error {
```
rfjakob: Needs a comment explaining the arguments

internal/nametransform/diriv.go
```diff
@@ -75,16 +75,17 @@ func fdReadDirIV(fd *os.File) (iv []byte, err error) {
 // WriteDirIV - create diriv file inside "dir" (absolute ciphertext path)
 // This function is exported because it is used from pathfs_frontend, main,
 // and also the automated tests.
-func WriteDirIV(dir string) error {
+func WriteDirIV(dirfd *os.File, dir string) error {
```
rfjakob: Function comment needs to be updated and dirfd explained.

Question: If dirfd != nil, can "dir" contain slashes? Should we check + panic?

internal/fusefrontend/fs_dir.go
slackner: OK

internal/nametransform/diriv.go
slackner: OK

It wouldn't be a problem if it contains slashes. Not sure if we should be too strict here.

internal/syscallcompat/sys_linux.go
slackner: The syscall itself? Or do you mean tests which use Rmdir in a FUSE mount?

internal/nametransform/diriv.go
rfjakob: It would work because the syscall accepts it, but is it a bug in gocryptfs if that happens?

internal/syscallcompat/sys_linux.go
rfjakob: Yes the syscall itself.

The Go stdlib uses https://golang.org/src/syscall/mksyscall.pl to autogenerate the wrappers, and different prototypes for amd64 ( https://golang.org/src/syscall/syscall_linux_amd64.go ) and the rest ( https://golang.org/src/syscall/syscall_linux.go ). Looks hard enough to get right (on 32 bit and 64 bit!) to warrant a test

rfjakob: (I will add a test for syscallat as well)

rfjakob:
I'm going to make the darwin syscalls testable on linux, and I don't want to cause merge conflicts, so I'll just pull this now. My nitpicks can be fixed later. Thanks!

slackner:
As I see you have already pushed your changes, looks awesome! :)
I will later open a new pull request for the comment fixes

PR #180, opened 2017-11-30
Even more syscallcompat improvements

slackner:
Patch 1 adds / updates the documentation as requested in https://github.com/rfjakob/gocryptfs/pull/179.

Patch 2 fixes the syscall emulation for absolute paths in a couple of functions. For absolute paths the `dirfd` should always be ignored. BTW, it would be nice if it would be possible to run the same emulation tests with emulation disabled, to test if the Linux syscalls actually work as expected.

Patch 3 updates two existing syscall wrappers with the new method (open current directory instead of caching the absolute path)

After this series only `emulateRenameat` remains a bit ugly. If you have any suggestions how to improve it without falling back to absolute paths ... ideas welcome ;)

### DIFF
--- a/internal/syscallcompat/emulate_test.go
+++ b/internal/syscallcompat/emulate_test.go
@@ -67,6 +67,15 @@ func TestEmulateOpenat(t *testing.T) {
 	if rawFd < 0 {
 		t.Fatalf("rawFd=%d", rawFd)
 	}
+	// Test with absolute path
+	rawFd, err = emulateOpenat(-1, tmpDir + "/testOpenAt", 0, 0)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer syscall.Close(rawFd)
+	if rawFd < 0 {
+		t.Fatalf("rawFd=%d", rawFd)
+	}
 }
 
 func TestEmulateRenameat(t *testing.T) {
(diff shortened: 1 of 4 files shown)


internal/syscallcompat/emulate_test.go
```diff
@@ -95,6 +104,15 @@ func TestEmulateRenameat(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// Test with absolute path
+	err = emulateRenameat(-1, tmpDir + "/dir2/f2", -1, tmpDir + "/dir2/f1")
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = os.Stat(tmpDir + "/dir2/f1")
```
slackner: Oh, small typo here.

slackner: Or not ... sorry for the noise (thought the path was incorrect, but should be fine - otherwise the tests also wouldn't pass)

rfjakob:
Looks good to me, thanks! I'll think about emulateRenameat.

rfjakob:
Oh and about testing the linux syscalls - that would be nice indeed. I have not found a solution for how to do that without massive copy+paste, though.

PR #185, opened 2017-12-07
Use openBackingDir in Readlink and GetAttr

slackner:
Also fixes another possible FD leak


rfjakob:
Looks good to me, thanks!

PR #186, opened 2017-12-08
fusefrontend_reverse: Use O_DIRECTORY in OpenDir implementation

slackner:
Also get rid of the defer - it is not really necessary here.

By the way: Do we also want to use the helper function here? It is not really suitable because we want only have a directory component.

### DIFF
--- a/internal/fusefrontend_reverse/rfs.go
+++ b/internal/fusefrontend_reverse/rfs.go
@@ -255,12 +255,12 @@ func (rfs *ReverseFS) OpenDir(cipherPath string, context *fuse.Context) ([]fuse.
 		return nil, fuse.ToStatus(err)
 	}
 	// Read plaintext dir
-	fd, err := syscallcompat.OpenNofollow(rfs.args.Cipherdir, relPath, syscall.O_RDONLY, 0)
+	fd, err := syscallcompat.OpenNofollow(rfs.args.Cipherdir, relPath, syscall.O_RDONLY|syscall.O_DIRECTORY, 0)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


rfjakob:
How about
```
openBackingDir(relPath + "/.")
```
?

slackner:
I thought about that, but unfortunately the same trick can not be used in `rfile.go:newFile` (different flags):

```
	pRelPath, err := rfs.decryptPath(relPath)
	if err != nil {
		return nil, fuse.ToStatus(err)
	}
	fd, err := syscallcompat.OpenNofollow(rfs.args.Cipherdir, pRelPath, syscall.O_RDONLY, 0)
	if err != nil {
		return nil, fuse.ToStatus(err)
	}
```

This means we either still need a second function, or have to keep a couple of places which don't use `openBackingPath`. Should we still use it in the `OpenDir` implementation?

rfjakob:
newFile could use openBackingDir + openat

rfjakob:
(not sure if that makes the code nicer)

In any case I would use the helper for openDir.

slackner:
You had almost convinced me, but after playing around a bit with it, I am not so sure anymore. ;)

Just appending "/." causes problems with the root directory (attempt to decode empty input). Using `filepath.Join(relPath, ".")` also doesn't append a separate component - the `Join` implementation just skips the useless path component. Even if we manually handle the root directory and translate it to "./." then we are back at the decoding error - the ciphertext also isn't expected to contain "." at this point.

Still worth the effort?

rfjakob:
Ok, let's keep it as is. Way too much trouble.

Merging this, thanks!

Issue #187, opened 2017-12-08
Race-condition when reading file content can access device nodes in reverse mode


slackner:
The general idea is similar to issue #165 that is already mostly fixed.

**Steps to reproduce:**

1. Create a regular reverse mount point
2. Create a file "test" in the original directory
3. Access the corresponding encrypted directory in the mount point (`ls <encrypted dir>`)
4. Quickly delete the file in the original data - instead create a device node
5. Access the file again, it will access the device node and attempt to read from it

In practice the impact of this possible attack is relatively small because step 4 already requires root permissions. Nevertheless, if I don't miss anything the problem can be easily avoided by checking the file type using the handle after the `OpenNofollow` call. Actually, there already is an `Fstat` call in the corresponding function.

PR #188, opened 2017-12-11
fusefrontend_reverse: Reject access to device nodes in newFile function

slackner:
**Note:** I am currently using `FromStat` to convert the syscall result to a `fuse.Attr` sruct and then `IsRegular()` to check if the handle refers to a regular file. If you prefer a different method (for example, manually checking the bits in the struct) please let me know.

Steps to reproduce:

* Create a regular reverse mount point
* Create a file "test" in the original directory
* Access the corresponding encrypted directory in the mount point (ls <encrypted dir>)
* Quickly delete the file in the original data - instead create a device node
* Access the file again, it will access the device node and attempt to read from it

Fixes https://github.com/rfjakob/gocryptfs/issues/187


rfjakob:
Works for me. Merging, thanks!

PR #189, opened 2017-12-12
Handle PlaintextNames mode in Link

slackner:
See https://github.com/rfjakob/gocryptfs/issues/174.

The second patch changes the code to use the use the more secure `Linkat` syscall.


rfjakob:
Merged, thanks!

PR #193, opened 2018-01-09
Provide basic man page for xray tool

lechner:
From Debian packaging. Couldn't find a manpage for the xray tool. Thank you!

### DIFF
--- a/Documentation/MANPAGE-XRAY.md
+++ b/Documentation/MANPAGE-XRAY.md
@@ -5,7 +5,7 @@
 NAME
 ====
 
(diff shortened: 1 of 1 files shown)


rfjakob:
Good stuff, thanks.

Two nitpicks:

1) "GoCryptfs" should be written as "gocryptfs"
2) Remove "fallocate(2)" from SEE ALSO, as it does not matter for gocryptfs-xray?

lechner:
Done. Thank you!

rfjakob:
Merged, thanks!

PR #204, opened 2018-02-03
Globally replace 'syscall' with 'golang.org/x/sys/unix'

lechner:
This might save you some time should you ever switch entirely to `golang.org/x/sys/unix`. The older `syscall` is still needed in a few places until Go-Fuse offers an API also based on `unix`.

### DIFF
--- a/internal/fusefrontend/file.go
+++ b/internal/fusefrontend/file.go
@@ -8,7 +8,7 @@ import (
 	"log"
 	"os"
 	"sync"
-	"syscall"
+	"golang.org/x/sys/unix"
 	"time"
 
 	"github.com/hanwen/go-fuse/fuse"
@@ -60,8 +60,8 @@ type file struct {
 
 // NewFile returns a new go-fuse File instance.
 func NewFile(fd *os.File, fs *FS) (nodefs.File, fuse.Status) {
-	var st syscall.Stat_t
-	err := syscall.Fstat(int(fd.Fd()), &st)
+	var st unix.Stat_t
+	err := unix.Fstat(int(fd.Fd()), &st)
 	if err != nil {
 		tlog.Warn.Printf("NewFile: Fstat on fd %d failed: %v\n", fd.Fd(), err)
 		return nil, fuse.ToStatus(err)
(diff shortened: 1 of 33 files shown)


internal/fusefrontend/fs.go
```diff
@@ -121,12 +120,12 @@ func (fs *FS) Open(path string, flags uint32, context *fuse.Context) (fuseFile n
 	f, err := os.OpenFile(cPath, newFlags, 0)
 	if err != nil {
 		sysErr := err.(*os.PathError).Err
-		if sysErr == syscall.EMFILE {
-			var lim syscall.Rlimit
-			syscall.Getrlimit(syscall.RLIMIT_NOFILE, &lim)
+		if sysErr == unix.EMFILE {
```
rfjakob: Hmm, this is a bit ugly. The error is returned by `os.OpenFile`, which uses the syscall package internally. But then we compare against the error definition in the unix package...

internal/fusefrontend/file.go
```diff
@@ -60,8 +60,8 @@ type file struct {
 
 // NewFile returns a new go-fuse File instance.
 func NewFile(fd *os.File, fs *FS) (nodefs.File, fuse.Status) {
-	var st syscall.Stat_t
-	err := syscall.Fstat(int(fd.Fd()), &st)
+	var st unix.Stat_t
+	err := unix.Fstat(int(fd.Fd()), &st)
 	if err != nil {
 		tlog.Warn.Printf("NewFile: Fstat on fd %d failed: %v\n", fd.Fd(), err)
 		return nil, fuse.ToStatus(err)
```
rfjakob: This may also be a problem - `fuse.ToStatus` compares the error it gets against `syscall` and `os` constants - it even checks the type here:
https://github.com/hanwen/go-fuse/blob/9060525339d688b3e8313f25eee96690d7314a60/fuse/misc.go#L54

	switch t := err.(type) {

rfjakob:
I'll not merge this as I believe the costs outweigh the benefit

rfjakob:
Note to self: sync.map needs Go 1.9 and higher: https://golang.org/doc/go1.9#sync-map

PR #206, opened 2018-02-04
Drop dependency on golang.org/x/syncmap

lechner:
For Debian, it was desirable to remove the dependency on `golang.org/x/syncmap` (now available in the standard library as `sync/map`). There is no need to merge this patch. It can be maintained indefinitely should upstream wish to remain compatible with older versions of Go.

P.S. There is another mention in `Gopkg.lock` but I do not know how to use or modify that file. Thank you!

### DIFF
--- a/internal/fusefrontend_reverse/rfile.go
+++ b/internal/fusefrontend_reverse/rfile.go
@@ -5,9 +5,7 @@ import (
 	"io"
 	"os"
 	"syscall"
-
-	// In newer Go versions, this has moved to just "sync/syncmap".
-	"golang.org/x/sync/syncmap"
+	"sync"
 
 	"github.com/hanwen/go-fuse/fuse"
 	"github.com/hanwen/go-fuse/fuse/nodefs"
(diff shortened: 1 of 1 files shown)


rfjakob:
Travis runs a pretty old Ubuntu 14.04 "Trusty" which has gccgo 4.9 ( https://packages.ubuntu.com/trusty/gccgo )

gocryptfs needs Go 1.5 or higher, and acc. to https://golang.org/doc/install/gccgo#Releases this means gccgo 6:

> The GCC 6 releases include a complete implementation of the Go 1.6.1 user libraries. The Go 1.6 runtime is not fully merged, but that should not be visible to Go programs.

rfjakob:
gccgo-6 is available via an Ubuntu repo, I have added it in https://github.com/rfjakob/gocryptfs/commit/88ed4140e5b9b6dc1df2f69cbe094630f5488941 , but it fails for some reason: https://travis-ci.org/rfjakob/gocryptfs/jobs/337760519#L712 :
> profiling.go:6:2: cannot find package "runtime/trace" in any of:

rfjakob:
Closing this to keep compatibility with gcc-go and older Go

rfjakob:
Posted this in the wrong ticket, this is the right one:

Note to self: sync.Map needs Go 1.9 and higher: https://golang.org/doc/go1.9#sync-map

PR #227, opened 2018-04-15
Improved xattr handling on non-linux systems

bolshevik:
- Removed xattr filtering for non-Linux. "system." and "user." prefixes are only relevant for Linux.
- Small cleanup and additional tests.

Disallowing xattribute name like `com.apple` on Darwin results in creation of `._` files to store the respective attributes.

### DIFF
--- a/internal/fusefrontend/xattr.go
+++ b/internal/fusefrontend/xattr.go
@@ -5,7 +5,6 @@ package fusefrontend
 
 import (
 	"strings"
-	"syscall"
 
 	"github.com/hanwen/go-fuse/fuse"
 	"github.com/pkg/xattr"
@@ -15,11 +14,6 @@ import (
 
 // xattr names are encrypted like file names, but with a fixed IV.
 var xattrNameIV = []byte("xattr_name_iv_xx")
-
-// Only allow the "user" namespace, block "trusted" and "security", as
-// these may be interpreted by the system, and we don't want to cause
-// trouble with our encrypted garbage.
-var xattrUserPrefix = "user."
 
 // We store encrypted xattrs under this prefix plus the base64-encoded
 // encrypted original name.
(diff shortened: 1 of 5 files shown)


internal/fusefrontend/xattr.go
```diff
@@ -124,21 +116,13 @@ func (fs *FS) ListXAttr(path string, context *fuse.Context) ([]string, fuse.Stat
...
 	cAttr = xattrStorePrefix + fs.nameTransform.EncryptName(attr, xattrNameIV)
-	return cAttr, nil
+	return cAttr
 }
 
 func (fs *FS) decryptXattrName(cAttr string) (attr string, err error) {
-	// Reject anything that does not start with "user.gocryptfs."
-	if !strings.HasPrefix(cAttr, xattrStorePrefix) {
-		return "", syscall.EINVAL
-	}
```
rfjakob: I think we want to keep this

bolshevik: Thanks! I returned it back, but probably this part can be also optimised even more, there is a code here https://github.com/rfjakob/gocryptfs/blob/master/internal/fusefrontend/xattr.go#L112 , which guarantees, that the method will not be called with a wrong parameter. What do you think, what is the best to prevent duplication of these checks?

rfjakob: Good point. I'm traveling right now, will think about it and merge in the next days.

rfjakob:
About the redundant prefix check in `decryptXattrName`: I think there are two options:

1) Document that the function may crash when the passed string is too short
2) Keep the check and return an error

In this case I think keeping the check is better because the check is cheap and it makes the life for future callers easier. Also, if a function has an error return value (and `decryptXattrName`) I would not expect that you can crash it with invalid input.

rfjakob:
Merged to master, thanks!

Issue #229, opened 2018-04-28
macOS: modification and creation dates reset


alexanderharm:
I started using `gocryptfs` (latest version) on macOS High Sierra and noticed that each time I eject my encrypted volume and remount it, that the creation and modification dates are reset to `1970-01-01`. Is this expected behaviour?

I mount my dirs using

```bash
gocryptfs -ko local /path/to/encDir /Volumes/encDir
```

Also mounting them without the `-ko local` option has the same result.

rfjakob:
Huh. What happens to the dates of the encrypted files?

alexanderharm:
The date of the metadata file (?) is correct, the date of the data file is also reset.

<img width="675" alt="screen shot 2018-04-28 at 21 40 39" src="https://user-images.githubusercontent.com/16108584/39400239-eece996a-4b2c-11e8-86a9-94709999bb76.png">

rfjakob:
When you write to a file, the encrypted file stays at 1970?

alexanderharm:
For a very brief moment it shows the correct datetime but after max. one second it is reset.

rfjakob:
Well, that's funny. I'll see if I can get my hands on a macos box next week to reproduce this.

I the meantime, can you run gocryptfs with

"-fg -nosyslog -fusedebug"

an see what happens when you modify a file?

alexanderharm:
I don't see anything unusual but then I don't have a clue about this :-)
[gocryptfs.log](https://github.com/rfjakob/gocryptfs/files/1958135/gocryptfs.log)

rfjakob:
Maybe found it. Can you try this build? [gocryptfs.gz](https://github.com/rfjakob/gocryptfs/files/1959095/gocryptfs.gz)
gocryptfs v1.4.4-29-g996d2f1 without_openssl; go-fuse v20170619-32-g4ff4497; 2018-04-29 go1.10

alexanderharm:
I tried and noticed the following:

* Mounted decrypted new file: creation date still 1970, modification date: fine

<img width="253" alt="screen shot 2018-04-29 at 21 36 16" src="https://user-images.githubusercontent.com/16108584/39410252-69a3502c-4bf5-11e8-86c2-c92004a2c869.png">

* Encrypted new file: both dates seem to look fine

* ***BUT***: I cannot modify existing files anymore (see attached screenshot)

[...]

PR #230, opened 2018-05-01
Xattr: Darwin and Apple native applications compatibility improvement 

bolshevik:
xattr: added passing of a "flags" parameter to the lower layer syscall. This makes Apple applications being able to successfully save files, otherwise it fails with a flags value != 0.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -4,20 +4,28 @@
 [[projects]]
   branch = "master"
   name = "github.com/hanwen/go-fuse"
-  packages = ["fuse","fuse/nodefs","fuse/pathfs","splice"]
+  packages = [
+    "fuse",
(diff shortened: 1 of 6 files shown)


rfjakob:
Very nice, thanks. Sounds like it could fix the save file issue in #229

bolshevik:
Yes, but here we should wait for a release of https://github.com/pkg/xattr/pull/26

rfjakob:
Merged to master as https://github.com/rfjakob/gocryptfs/commit/95964fb5f0646ffb21d7b67ac9077ba2ef880526 and https://github.com/rfjakob/gocryptfs/commit/5ccc06d5cb321b1400e33a7d0819bf617c52600d ,
thanks!

PR #233, opened 2018-05-06
Fix some problems with /etc/fstab

mahkoh:
These patches fix some problems I encountered while using gocryptfs via /etc/fstab. See the commit messages for more details.


rfjakob:
Hi, thanks for the changes!

About -noexec, I wonder if we should actually mount with -noexec instead of ignoring it.

rfjakob:
Merged with some modifications as https://github.com/rfjakob/gocryptfs/commit/10212d791a3196c2c8705a7a3cccdeb14a8efdbe and https://github.com/rfjakob/gocryptfs/commit/53d6a9999dd0e4c31636d16179f284fff35a35d9 - thanks!

PR #234, opened 2018-05-07
xattr: removed base64-encoding of values

bolshevik:
Added decryptAsFirstBlock and encryptAsFirstBlock methods to FS in order to optimize xattr storage: values a binary-safe, there is no need to base64-encode them.

**Backward compatibility breaking change!** One needs to migrate the attributes of the old files to the new version. For example:
1) Mount the same directory with the previous and this version simultaneously.
2) Execute `rsync -aX OLD_MOUNT/ NEW_MOUNT/`
3) Ignore I/O errors.
4) Unmount the folders and update `gocryptfs` binary.

### DIFF
--- a/internal/fusefrontend/fs.go
+++ b/internal/fusefrontend/fs.go
@@ -351,6 +351,11 @@ func (fs *FS) StatFs(path string) *fuse.StatfsOut {
 	return fs.FileSystem.StatFs(cPath)
 }
 
+// decryptAsFirstBlock: decrypts data as at were 0 block of the file (encrypted with GCM)
+func (fs *FS) decryptAsFirstBlock(block []byte) ([]byte, error) {
+	return fs.contentEnc.DecryptBlock(block, 0, nil)
+}
+
 // decryptSymlinkTarget: "cData64" is base64-decoded and decrypted
 // like file contents (GCM).
 // The empty string decrypts to the empty string.
@@ -362,7 +367,7 @@ func (fs *FS) decryptSymlinkTarget(cData64 string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	data, err := fs.contentEnc.DecryptBlock([]byte(cData), 0, nil)
+	data, err := fs.decryptAsFirstBlock([]byte(cData))
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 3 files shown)


rfjakob:
Not sure about this - even though xattr values are binary-safe, most xattrs in the Linux world are ASCII as far as I know (think SELinux attributes). Encoding to base64 seems to be the "safe" option, downside is 30% size increase, but does that matter?

bolshevik:
Thanks for your feedback!

I think better would be not to have any assumption, check also the `getfattr -e`, actually it is encoding to base64 by default for you the same way one can export in other encodings if needed. 

Regarding SELinux: will it in general work? I thought there were problems accessing FUSE filesystems from the kernel, weren't it?

Or check the dropbox sync client, it adds binary meta-data into attributes (probably hash or whatever).
On the MacOS the tools also embed binary data for flags and other data. The increase of size is indeed 33,2%, but you know, why should we steal from limits of the OS or underlying FS system if we can avoid it.

rfjakob:
I have actually no idea if SELinux works over FUSE, I just notices that SELinux uses plain-text xattr values.

But indeed, binary xattrs are actually more common than I thought. Question is, how many people have used gocryptfs since commit https://github.com/rfjakob/gocryptfs/commit/8b443c8484f5fdcedbec2a689a7d66d2a277b26e (1 month ago). Changing the format now means reading attributes in the old format will throw an I/O error.

bolshevik:
@rfjakob Thanks for the statistics, I think we should choose one of these:
1) BC-Release with a migration path as above.

2) We can add a new prefix `user.gocryptfsb.` for writing and reading by default and have the new logic enabled, when the attribute does not exist, we can fall back to the `user.gocryptfs.` but only for reading.

The `user.gocryptfs.` can be deprecated and removed after some time. But here in the worst case we leave garbage data.

3) We can build "ifs" in the reading function and make it fallback to base64 to make it transparent and without needing to rename the prefix. But next writing will overwrite the data in the new format. 

Point 3 is discouraging migration, maybe we can also ship a small shell script or program for 2 and 3, that traverses all files and rewrites the extended attributes.

What would we choose?

rfjakob:
I think we should either keep base64 as-is, or do (3), like this:

1) Try to decrypt. If that fails, log a Warning. Then,
2) Try to base64-decode
3) Try to decrypt again

This also needs a testcase to verify both formats work (maybe similar to https://github.com/rfjakob/gocryptfs/blob/master/tests/fsck/fsck_test.go#L15, but should be in https://github.com/rfjakob/gocryptfs/tree/master/tests/xattr).

bolshevik:
I've added changes according to solution number 3. Please review.

Nodens-:
As someone who is beyond https://github.com/rfjakob/gocryptfs/commit/8b443c8484f5fdcedbec2a689a7d66d2a277b26e for quite a while now and also has several terabytes of data on slowish remote filesystems (which make migration via copy very very tedious), the transparent option sounds great.

Regarding logging a warning on fail, why not just log after base64 has failed as well only? I see no purpose for the extra log there.

rfjakob:
Slightly edited and merged as https://github.com/rfjakob/gocryptfs/commit/a41ec2028cf2ad24eafb531d661ea1fe0f751f39 , thanks!

@Nodens- thanks for the feedback, extra warning is gone.

bolshevik:
@rfjakob @Nodens- Thank you!

PR #265, opened 2018-10-06
Added option for autounmount

duncanka:
Even though filesystem notifications aren't implemented for FUSE, I decided to try my hand at implementing the autounmount feature (#128). I based it on the EncFS [autounmount code](https://github.com/vgough/encfs/blob/1974b417af189a41ffae4c6feb011d2a0498e437/encfs/main.cpp#L851), which records filesystem accesses and checks every 10 seconds whether it's idled long enough to unmount.

I've tested the feature locally, but I haven't added any tests for this flag. I also haven't worked with Go before. So please let me know if there's anything that should be done differently.

One particular concern: I worked from the assumption that the open files table is unique per-filesystem. If that's not true, I'll need to add an open file count and associated lock to the Filesystem type instead.

### DIFF
--- a/cli_args.go
+++ b/cli_args.go
@@ -33,6 +33,8 @@ type argContainer struct {
 	// Configuration file name override
 	config             string
 	notifypid, scryptn int
+	// Autounmount
+	autounmount int
 	// Helper variables that are NOT cli options all start with an underscore
 	// _configCustom is true when the user sets a custom config file name.
 	_configCustom bool
@@ -187,6 +189,9 @@ func parseCliOpts() (args argContainer) {
 		"successful mount - used internally for daemonization")
 	flagSet.IntVar(&args.scryptn, "scryptn", configfile.ScryptDefaultLogN, "scrypt cost parameter logN. Possible values: 10-28. "+
 		"A lower value speeds up mounting and reduces its memory needs, but makes the password susceptible to brute-force attacks")
+
+	flagSet.IntVar(&args.autounmount, "autounmount", 0, "Idle time in minutes before autounmount (forward mode only). 0 means stay mounted indefinitely.")
+
 	var dummyString string
 	flagSet.StringVar(&dummyString, "o", "", "For compatibility with mount(1), options can be also passed as a comma-separated list to -o on the end.")
 	// Actual parsing
--- a/internal/fusefrontend/fs.go
+++ b/internal/fusefrontend/fs.go
@@ -47,6 +47,15 @@ type FS struct {
 	// "gocryptfs -fsck" reads from the channel to also catch these transparently-
 	// mitigated corruptions.
 	MitigatedCorruptions chan string
+	// Track accesses to the filesystem so that we can know when to autounmount.
+	// An access is considered to have happened on every call to encryptPath,
+	// which is called as part of every filesystem operation.
+	// (This variable shouldn't need a lock, as all that ever happens to it is that
+	// it gets set by an accessing thread or read and reset by the monitor thread.
+	// Races may result in a read of false while another thread is recording an access,
+	// but it's equally reasonable under those circumstances to let the read win
+	// and autounmount as it is to reset the idle monitor.)
+	AccessedSinceLastCheck bool
 }
 
 var _ pathfs.FileSystem = &FS{} // Verify that interface is implemented.
@@ -60,10 +69,11 @@ func NewFS(args Args, c *contentenc.ContentEnc, n *nametransform.NameTransform) 
 		tlog.Warn.Printf("Forward mode does not support -exclude")
 	}
 	return &FS{
-		FileSystem:    pathfs.NewLoopbackFileSystem(args.Cipherdir),
-		args:          args,
-		nameTransform: n,
-		contentEnc:    c,
+		FileSystem:             pathfs.NewLoopbackFileSystem(args.Cipherdir),
+		args:                   args,
+		nameTransform:          n,
+		contentEnc:             c,
+		AccessedSinceLastCheck: false,
 	}
 }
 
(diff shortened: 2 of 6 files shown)


rfjakob:
Hi and thanks for the pr!

Two comments for now:

1) I think we should use "-i" like encfs does. Let"s keep conventions if possible

2) Please flagSet.DurationVar . This allows passing strings like "20m" or "24h".

internal/fusefrontend/fs.go
```diff
@@ -60,10 +69,11 @@ func NewFS(args Args, c *contentenc.ContentEnc, n *nametransform.NameTransform)
...
 	return &FS{
-		FileSystem:    pathfs.NewLoopbackFileSystem(args.Cipherdir),
-		args:          args,
-		nameTransform: n,
-		contentEnc:    c,
+		FileSystem:             pathfs.NewLoopbackFileSystem(args.Cipherdir),
+		args:                   args,
+		nameTransform:          n,
+		contentEnc:             c,
+		AccessedSinceLastCheck: false,
```
rfjakob: You can drop this - booleans are initialized to false by default

cli_args.go
```diff
@@ -33,6 +33,8 @@ type argContainer struct {
 	// Configuration file name override
 	config             string
 	notifypid, scryptn int
+	// Autounmount
+	autounmount int
```
rfjakob: Please just call it "idle", like EncFS does

cli_args.go
```diff
@@ -187,6 +189,9 @@ func parseCliOpts() (args argContainer) {
 		"successful mount - used internally for daemonization")
 	flagSet.IntVar(&args.scryptn, "scryptn", configfile.ScryptDefaultLogN, "scrypt cost parameter logN. Possible values: 10-28. "+
 		"A lower value speeds up mounting and reduces its memory needs, but makes the password susceptible to brute-force attacks")
+
+	flagSet.IntVar(&args.autounmount, "autounmount", 0, "Idle time in minutes before autounmount (forward mode only). 0 means stay mounted indefinitely.")
```
rfjakob: Please use `flagSet.DurationVar`

help.go
```diff
@@ -19,6 +19,7 @@ func helpShort() {
 Common Options (use -hh to show all):
   -aessiv            Use AES-SIV encryption (with -init)
   -allow_other       Allow other users to access the mount
+  -autounmount       Unmount automatically after n minutes of idle time
```
rfjakob: "-idle"

mount.go
```diff
@@ -137,10 +138,49 @@ func doMount(args *argContainer) {
...
+		fwdFs := fs.(*fusefrontend.FS)
+		go idleMonitor(args.autounmount, fwdFs, srv, args.mountpoint)
+	}
 	// Jump into server loop. Returns when it gets an umount request from the kernel.
 	srv.Serve()
 }
 
+// Based on the EncFS idle monitor:
+// https://github.com/vgough/encfs/blob/1974b417af189a41ffae4c6feb011d2a0498e437/encfs/main.cpp#L851
+const ActivityCheckIntervalSeconds = 10
```
rfjakob: Can you use a fraction of idleTimeout instead? For automated tests, I would like to use sub-second timeouts (which is also why I want to use flagSet.DurationVar instead of minutes)

mount.go
```diff
@@ -137,10 +138,49 @@ func doMount(args *argContainer) {
...
+// idleMonitor is a function to be run as a thread that checks for
+// filesystem idleness and unmounts if we've been idle for long enough.
+func idleMonitor(idleTimeout int, fs *fusefrontend.FS, srv *fuse.Server, mountpoint string) {
+	timeoutCycles := 60 * idleTimeout / ActivityCheckIntervalSeconds
+	idleCount := 0
+	for true {
```
rfjakob: In Go, the infinite loop is written like
```
for {

}
```
(omit "true")

mount.go
duncanka: Hmm...one potential issue there is that we could end up waiting up to idleTimeout / checkIntervalFraction extra if an access occurs just after we check for idleness. E.g., if the timeout is 10 hours and we check 5 times during the timeout period, an access right after we check for idle would make us consider the next 2 hours to be busy. Maybe idleTimeout / 4 with a cap at 2 minutes?

mount.go
```diff
@@ -137,10 +138,49 @@ func doMount(args *argContainer) {
...
+		recentAccess := fs.AccessedSinceLastCheck
+		// Don't worry about flag being set here by another thread...
+		// on the scale of minutes, it might as well have happened
+		// just before we checked.
+		fs.AccessedSinceLastCheck = false
```
duncanka: Just to be safe, should we use CompareAndSwapUint32 from the atomic package here? This is literally just a compare-and-swap operation, after all.

duncanka:
I have no idea why the Travis checks are failing...some constant is overflowing int, but I don't see how any of the changes in 185d05e could have led to that.

mount.go
rfjakob: Yes!

rfjakob:
`constant 4294967295 overflows int` failures are from an unrelated change in the go-fuse library ( https://github.com/hanwen/go-fuse/pull/237 ). We can ignore that here.

mount.go
rfjakob: By the way, I think integer math would be good enough here - time.Duration values are in nanoseconds

duncanka:
I just added a commit to the branch in my copy of the repo, but it's not showing up here, presumably because it's marked as closed?

rfjakob:
No problem, i can pull directly. However, the change https://github.com/duncanka/gocryptfs/commit/30d4bd329bc40154ecd0b422b7198997e37a7e03 strips the trailing double-spaces, which are line breaks in markdown!

duncanka:
Whoops, sorry! Replaced with a trailing-space-preserving version.

rfjakob:
Merged as https://github.com/rfjakob/gocryptfs/commit/0751b4e00a3cbb7e940e39d76abadd21aadcd16c , thanks!

PR #292, opened 2018-12-26
configfile: Explicitly wipe scrypt derived key after decrypting/encrypting master key.

slackner:
Further raises the bar for recovering keys from memory.

The `scryptHash` array was already cleared in `EncryptKey` (added in 0efd220d1e10ac8e3d0048ff4d068cc8174e7185). However, the same logic is also required in `DecryptMasterKey`. In addition, there might be copies of the key stored in the cryptocore instance, so explicitly call `Wipe()` (using a returned callback method).

### DIFF
--- a/internal/configfile/config_file.go
+++ b/internal/configfile/config_file.go
@@ -224,6 +224,15 @@ func (cf *ConfFile) DecryptMasterKey(password []byte) (masterkey []byte, err err
 	tlog.Warn.Enabled = false // Silence DecryptBlock() error messages on incorrect password
 	masterkey, err = ce.DecryptBlock(cf.EncryptedKey, 0, nil)
 	tlog.Warn.Enabled = true
+
+	// Purge scrypt-derived key
+	for i := range scryptHash {
+		scryptHash[i] = 0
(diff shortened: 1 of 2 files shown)


rfjakob:
Thanks, I agree with the intention, but i think that instead of returning a callback, ContentEnc should get a Wipe() method.

slackner:
I've updated my branch based on your feedback. Is this what you hand in mind?

rfjakob:
Yes exactly, thanks!

My greatest fear is accidentially using an all-zero key for encryption, can you

1) set "scryptHash = nil" after overwriting with zeros in both locations
2) set "ce = nil" after calling Wipe() in both locations
3) set "be.cryptoCore = nil" after calling Wipe()

rfjakob:
This should make sure that any code (current or future) that tries to use the wiped keys crashes instead of using all-zero encryption.

slackner:
I've updated my commit based on your feedback. Yes, it certainly makes sense to be extra careful here. ;)

rfjakob:
Perfect, thanks!

PR #301, opened 2018-12-29
nametransform: Delete incomplete longname files on error.

slackner:
Handle errors in the same way as in `WriteDirIV`.
Should we somehow unify those two functions?


rfjakob:
Merged, thanks. Yes this logic could be moved to a helper function in syscallcompat, `WriteComplete` maybe?

PR #310, opened 2019-01-03
Symlink race fixes

slackner:
Several fixes for the `symlink_race` branch.

### DIFF
--- a/init_dir.go
+++ b/init_dir.go
@@ -99,7 +99,7 @@ func initDir(args *argContainer) {
 	// in the root dir
 	if !args.plaintextnames && !args.reverse {
 		// Open cipherdir (following symlinks)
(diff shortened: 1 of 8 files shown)


slackner:
Actually, before you merge it, I have some more changes. I will update this branch in a few minutes.

slackner:
Okay, updated! Please let me know if you are fine with these changes. (If you already did some of them yourself, that is also fine ofc.)

rfjakob:
Looks good, thanks!

slackner:
No problem. By the way: What is the reason for the new errors in the tests? They are not introduced by these commits, but a lot of tests fail with:

```
fatal: No names found, cannot describe anything.
```

rfjakob:
This is because travis only clones the last 100 commits, and when we are more than 100 after the last tag, this happens. Just pushed a fix by bumping to 300.

Issue #305, opened 2019-01-01
Truncate does not create sparse file of size 4096


slackner:
In order to reproduce this problem, run the following command in a gocryptfs mount point (the file should not exist yet or have a size of zero):

`truncate -s 4096 sparsefile`

The expected result would be that a sparse file is created, i.e., the file should consist of 18 bytes header and then 4096 zero bytes. However, for size 4096 this does not work. The result is a non-sparse file. Note that for other sizes everything works as expected.

The problem seems to be located in `internal/fusefrontend/file_allocate_truncate.go`. The condition `if n1 == n2 {` should be skipped if `oldPlainSz == 0`, otherwise `n1` is 0 and the code assumes the write goes to an existing block. My proposed patch would be something like this (mostly untested):

```
@@ -181,18 +184,17 @@ func (f *File) truncateGrowFile(oldPlainSz uint64, newPlainSz uint64) fuse.Statu
        if newPlainSz <= oldPlainSz {
                log.Panicf("BUG: newSize=%d <= oldSize=%d", newPlainSz, oldPlainSz)
        }
-       var n1 uint64
-       if oldPlainSz > 0 {
-               n1 = f.contentEnc.PlainOffToBlockNo(oldPlainSz - 1)
-       }
        newEOFOffset := newPlainSz - 1
-       n2 := f.contentEnc.PlainOffToBlockNo(newEOFOffset)
-       // The file is grown within one block, no need to pad anything.
-       // Write a single zero to the last byte and let doWrite figure out the RMW.
-       if n1 == n2 {
-               buf := make([]byte, 1)
-               _, status := f.doWrite(buf, int64(newEOFOffset))
-               return status
+       if oldPlainSz > 0 {
+               n1 := f.contentEnc.PlainOffToBlockNo(oldPlainSz - 1)
+               n2 := f.contentEnc.PlainOffToBlockNo(newEOFOffset)
+               // The file is grown within one block, no need to pad anything.
+               // Write a single zero to the last byte and let doWrite figure out the RMW.
+               if n1 == n2 {
+                       buf := make([]byte, 1)
+                       _, status := f.doWrite(buf, int64(newEOFOffset))
+                       return status
+               }
        }
        // The truncate creates at least one new block.
        //
```

PR #314, opened 2019-01-03
fusefrontend: Allow to create sparse file of size 4096.

slackner:
When the old size is zero, there are no existing blocks to merge the
new data with. Directly use Ftruncate if the size is block-aligned.

Fixes https://github.com/rfjakob/gocryptfs/issues/305


rfjakob:
Tested and looks good, thanks!

Issue #309, opened 2019-01-03
fsck fails when files do not have sufficient access permissions


slackner:
When files do not have sufficient access permissions (e.g., `chmod 000 xattrtest`), `fsck` thinks that those files are corrupt:

```
fsck: error reading xattr "user.foo" from "xattrtest": 13=permission denied
fsck: error opening file "xattrtest": 13=permission denied
fsck summary: 2 corrupt files
```

The files should either be skipped or temporarily `chmod`-ed to perform the file system check.

PR #315, opened 2019-01-04
fsck: Do not show files/directories without access permissions as corrupt.

slackner:
When running as a regular user, error EACCES does not necessarily mean that the
file/directory/xattr is corrupt, but just that we do not have sufficient access
permissions. Add a hint that running as root can be used to check everything.

Fixes: https://github.com/rfjakob/gocryptfs/issues/309

----

Note that I decided against using `chmod` since this is not necessarily sufficient to get access in a multi-user environment. Also, it is too risky to leave behind the file system in an inconsistent state when the user cancels `fsck`.

### DIFF
--- a/fsck.go
+++ b/fsck.go
@@ -22,18 +22,30 @@ type fsckObj struct {
 	fs *fusefrontend.FS
 	// List of corrupt files
 	corruptList []string
+	// List of skipped files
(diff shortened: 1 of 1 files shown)


rfjakob:
Works for me!

PR #318, opened 2019-01-05
A few more spelling fixes.

slackner:


slackner:
Hm, I am not really sure if we can automate that. The command is just `codespell ./` (to show the errors) and `codespell -w ./` to fix them. However, unfortunately the program also detects a lot of false-positives. It also wants to correct Go package names (`flate` -> `flat`) and even `*.md5sums` files. Should the makefile just show them (including all the false positives)?

PR #321, opened 2019-01-05
main: Run 'ensure fds' code early during the program startup.

slackner:
The files are apparently processed in alphabetic order, so cli_args.go is
processed before main.go. In order to run before the go-fuse imports, put
the 'ensure fds' code in a separate package. Debug messages are omitted
to avoid additional imports (that might contain other code messing up our
file descriptors).

---

Before merging, please verify that this also works for you.

### DIFF
--- a/cli_args.go
+++ b/cli_args.go
@@ -1,6 +1,9 @@
 package main
 
 import (
+	// Should be initialized before anything else.
+	_ "github.com/rfjakob/gocryptfs/internal/ensurestdfds"
+
 	"flag"
 	"fmt"
 	"net"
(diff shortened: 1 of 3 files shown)


slackner:
Another reason why it makes sense to delete the debug messages: There is no guarantee that they end up in a valid file descriptor anyway. ;)

slackner:
Instead of adding it to `cli_args.go`, another option might also be to add a new file with a name < `cli_args.go`. Then we can skip the `internal/...` package.

rfjakob:
I think we need the new package in any case, as otherwise the init functions of other packages run first

slackner:
You are right. Just having a file `aaaaaa.go` with some init code is not sufficient. :(

rfjakob:
Merged as https://github.com/rfjakob/gocryptfs/commit/7e05e809b7579ad0473fff6ce466c19d417d4e93 , thanks!

Issue #311, opened 2019-01-03
Allocate FUSE call reserves 18 bytes too much disk space


slackner:
Steps to reproduce:
```
fallocate -l 8192 testfile
```

The debug output shows:
```
Allocate off=0 sz=8192 mode=0 cipherOff=18 cipherSz=8274
```

A reservation of 8274 byte is expected (i.e., two blocks with 4092+32 byte, and a 18 byte header), however, since `cipherOff != 0`, this effectively reserves 18 bytes too much.

slackner:
I would also argue that allocating a region in the middle of the file should always reserve full blocks since partial blocks are not allowed there.

rfjakob:
You mean a reservation of 8266 bytes is expected?

I checked with strace, we reserve 8274 bytes, as you said:

> [pid  6423] ftruncate(14, 8274)         = 0

slackner:
The problem seems to be here:

```
	cipherSz := lastBlock.BlockCipherOff() - cipherOff +
		f.contentEnc.PlainSizeToCipherSize(lastBlock.Skip+lastBlock.Length)
```

The function `f.contentEnc.PlainSizeToCipherSize` also adds the 18 bytes header, but this would not be required here. I hope it is now more clear what I mean.

Nevertheless, even with that fixed, the function should behave differently if blocks are somewhere in the middle of the file. Reserving 1 byte of an inner block (of size 4096) should reserve the full block (since it will be required anyway as soon as we write to the file). Otherwise there is no guarantee that the write will succeed.

rfjakob:
Ah, here it is, missed it in the strace output:
> [pid  6423] fallocate(14, FALLOC_FL_KEEP_SIZE, 18, 8274) = 0

slackner:
Another corner case that is not handled correctly yet: If the previous file size is zero, we should also explicitly fallocate the first 18 bytes for the header.

rfjakob:
Merged as 40d2427fd7f788262dc3bf220f789b7bd5fe781b , thanks

PR #326, opened 2019-01-06
fusefrontend: Fix computation of cipherSz in Allocate FUSE call.

slackner:
Do not use PlainSizeToCipherSize() since this adds the 18 bytes file header.

Partially fixes https://github.com/rfjakob/gocryptfs/issues/311

### DIFF
--- a/internal/fusefrontend/file_allocate_truncate.go
+++ b/internal/fusefrontend/file_allocate_truncate.go
@@ -62,7 +62,7 @@ func (f *File) Allocate(off uint64, sz uint64, mode uint32) fuse.Status {
 	// the file.
 	cipherOff := firstBlock.BlockCipherOff()
 	cipherSz := lastBlock.BlockCipherOff() - cipherOff +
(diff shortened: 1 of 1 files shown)


slackner:
Note that this does not fix all the issues, but it is definitely better than before. The following two corner cases are still not correct:

* If the file previously had size zero, also fallocate space for the header (most likely happens anyway due to granularity restrictions of the underlying FS)
* Reserving an inner region of a file should always reserve full blocks (since blocks < 4096+32 byte are not allowed there).

rfjakob:
Looks ok but needs a regression test. If you want to add one please go ahead, otherwise I will do it in the eventing.

rfjakob:
PS: https://github.com/rfjakob/gocryptfs/blob/master/tests/matrix/matrix_test.go#L235

slackner:
I don't mind if you want to take care of it, I will not have time to work on it for the next few hours.

rfjakob:
Ok!

rfjakob:
Merged as 40d2427fd7f788262dc3bf220f789b7bd5fe781b

Issue #259, opened 2018-08-26
Fchmodat does not support AT_SYMLINK_NOFOLLOW


rfjakob:
These were silently ignored until now (!) but are rejected by Go 1.11 stdlib: https://github.com/golang/go/issues/20130 for more info.

From the fchmodat man page:

     AT_SYMLINK_NOFOLLOW ... This flag is not currently implemented.

@slackner Ideas?

slackner:
As I see you have already temporarily removed the flag, that is also what I would have suggested.
Do you think it makes sense to implement a workaround based on `openat` and then `fchmod`?

rfjakob:
Good idea, that would fix it. I'll benchmark the performance impact.

rfjakob:
Haha. Problem with `openat`: We cannot do it if we don't have read permissions for the file:
```
--- FAIL: TestChmod (0.00s)
    matrix_test.go:905: chmod 000 -> 111 failed: permission denied
    matrix_test.go:905: chmod 000 -> 123 failed: permission denied
    matrix_test.go:905: chmod 000 -> 321 failed: permission denied
```

Issue #327, opened 2019-01-06
Possible race-conditions between file creation and Fchownat


slackner:
The method would work as follows:

* Assume a system contains a gocryptfs mount as root user with `-allow_other`
* As a regular user create a new file with mode containing the SUID flag and write access for other users
* Before gocryptfs executes the `Fchownat` call, try to open the file again, write some exploit code to it, and try to run it.

For a short time, the file is owned by root and has the SUID flag, so this is pretty dangerous. To mitigate this issue, we could temporarily change the effective UID/GID and skip the `Fchown` step.

Some more thoughts:

* In `Open`, we might want to explicitly filter `O_CREAT` from the flags - if it is ever used for file creation, it skips the `PreserveOwner` and `LongName` creation code.

* The same issue might also be relevant for `Mknod`. Should a regular user really be able to create block devices with arbitrary modes? This probably needs further investigations.

rfjakob:
Good find! The problem with seteuid is that affects all threads. We could create the file with mode 000, chown it, and then chmod to the requested mode.

Filtering O_creat is a good idea.

About mknod, the kernel checks if the user can do that before sending us the request. User should get permission denied from the kernel

rfjakob:
Actually, i'm not sure if we can get o_create in open, but blindly filtering it out does no harm.

slackner:
I have been thinking about this issue a bit more. Apparently, it is possible to change the effective permissions for a single thread on Linux, this is just a limitation of the POSIX api.

See http://man7.org/linux/man-pages/man2/setresuid.2.html:

>       At the kernel level, user IDs and group IDs are a per-thread
>       attribute.  However, POSIX requires that all threads in a process
>       share the same credentials.  The NPTL threading implementation
>       handles the POSIX requirements by providing wrapper functions for the
>       various system calls that change process UIDs and GIDs.  These
>       wrapper functions (including those for setresuid() and setresgid())
>       employ a signal-based technique to ensure that when one thread
>       changes credentials, all of the other threads in the process also
>       change their credentials.  For details, see nptl(7).

[...]

Issue #307, opened 2019-01-02
Revisit macOS syscall emulation code


slackner:
As mentioned in https://github.com/rfjakob/gocryptfs/commit/b26080bf96b85179d7dec37cb10bda43b53fd7e8#r31812909, the `sys/unix` package now contains most of the syscalls required by gocryptfs even for macOS. This should allow to reduce the syscall emulation code (only a few functions are still required) and improve performance on macOS.

rfjakob:
Good idea, my problem with MacOS changes is that I don't have a machine to test

slackner:
Unfortunately I also cannot easily test on macOS. Maybe we should add the 'help wanted' tag, it would definitely be useful to get some more macOS developers involved in gocryptfs.

Another option might be to use Travis CI for the OSX tests. Recent documentations suggest that OSX support is available for open-source projects:

https://docs.travis-ci.com/user/reference/osx/
https://docs.travis-ci.com/user/multi-os/

`go` is explicitly listed as supported language.

rfjakob:
Travis may really work if we manage to install OSXFUSE:

> Passwordless sudo is available

rfjakob:
Looks like these people got it to work:

https://github.com/bazelbuild/sandboxfs/blob/master/admin/travis-install.sh#L30

slackner:
Okay, good and bad news. The good one: I now have a mac machine available for testing and debugging. The bad one: Current versions of gocryptfs are pretty much broken on macOS. In forward mode, mountpoints cannot be accessed at all - you always get `invalid argument`. The first few fixes seem pretty easy, so maybe I can get them done today. Nevertheless, to avoid unnecessary rebasing work, it would be good if we could soon come up with a decision regarding the `seteuid` stuff. ;)

slackner:
Remaining issues after the `Faccessat` fix:

```
$ ./test-without-openssl.bash 
Using TMPDIR=/var/folders/6x/tlqzyq0j38j9xw6qkl8p3jcm0000gq/T/
flock is not available, skipping
gocryptfs v1.7-rc1-31-g116938c without_openssl; go-fuse v20170619-77-g425e8d5; 2019-01-12 go1.11.4
ok  	github.com/rfjakob/gocryptfs	0.008s
?   	github.com/rfjakob/gocryptfs/gocryptfs-xray	[no test files]
ok  	github.com/rfjakob/gocryptfs/gocryptfs-xray/xray_tests	0.055s
[...]

PR #330, opened 2019-01-06
syscallcompat: Implement workaround for Fchmodat with AT_SYMLINK_NOFOLLOW.

slackner:
Fixes https://github.com/rfjakob/gocryptfs/issues/259

----

In this case we have to be a bit careful: Is the trick really supported on all versions of Linux?

### DIFF
--- a/internal/syscallcompat/sys_linux.go
+++ b/internal/syscallcompat/sys_linux.go
@@ -2,6 +2,7 @@
 package syscallcompat
 
 import (
+	"fmt"
 	"sync"
 	"syscall"
 
@@ -97,12 +98,43 @@ func Dup3(oldfd int, newfd int, flags int) (err error) {
 
 // Fchmodat syscall.
 func Fchmodat(dirfd int, path string, mode uint32, flags int) (err error) {
-	// Linux does not support passing flags to fchmodat! From the man page:
+	// Linux does not support passing flags to Fchmodat! From the man page:
 	// AT_SYMLINK_NOFOLLOW ... This flag is not currently implemented.
 	// Linux ignores any flags, but Go stdlib rejects them with EOPNOTSUPP starting
 	// with Go 1.11. See https://github.com/golang/go/issues/20130 for more info.
 	// TODO: Use fchmodat2 once available on Linux.
-	return syscall.Fchmodat(dirfd, path, mode, 0)
+
+	// Why would we ever want to call this without AT_SYMLINK_NOFOLLOW?
+	if flags&unix.AT_SYMLINK_NOFOLLOW == 0 {
+		tlog.Warn.Printf("Fchmodat: adding missing AT_SYMLINK_NOFOLLOW flag")
+	}
+
+	// Open handle to the filename (but without opening the actual file).
+	fd, err := syscall.Openat(dirfd, path, syscall.O_NOFOLLOW|O_PATH, 0)
+	if err != nil {
+		return err
+	}
+	defer syscall.Close(fd)
+
+	// Now we can check the type without the risk of race-conditions.
+	var st syscall.Stat_t
+	err = syscall.Fstat(fd, &st)
+	if err != nil {
+		return err
+	}
+
+	// Return syscall.ELOOP if path refers to a symlink.
+	var a fuse.Attr
+	a.FromStat(&st)
+	if a.IsSymlink() {
+		return syscall.ELOOP
+	}
+
+	// Change mode of the actual file. Note that we can neither use
+	// Fchmodat (since fd is not necessarily a directory) nor Fchmod
+	// (since we are using O_PATH).
+	procPath := fmt.Sprintf("/proc/self/fd/%d", fd)
+	return syscall.Chmod(procPath, mode)
 }
 
 // Fchownat syscall.


rfjakob:
Good question. Travis runs a really old kernel, so if Travis passes that's good enough I think.

Comment about the code: I don't see a reason why this should ever be called without NOFOLLOW, and the current code does not:

```
$ git grep syscallcompat.Fchmodat
internal/fusefrontend/fs.go:    err = syscallcompat.Fchmodat(dirfd, cName, mode, unix.AT_SYMLINK_NOFOLLOW)
internal/fusefrontend/fs_dir.go:                err = syscallcompat.Fchmodat(dirfd, cName, origMode, unix.AT_SYMLINK_NOFOLLOW)
internal/fusefrontend/fs_dir.go:                err = syscallcompat.Fchmodat(parentDirFd, cName, origMode|0700, unix.AT_SYMLINK_NOFOLLOW)
internal/fusefrontend/fs_dir.go:                                err = syscallcompat.Fchmodat(parentDirFd, cName, origMode, unix.AT_SYMLINK_NOFOLLOW)
```
I think we can drop the if, and maybe add a comment that NOFOLLOW is always used.

rfjakob:
I'll check what glibc does, but it seems to also use the /proc/self/fd trick, https://manned.org/fchmodat.2#head8 :
```
NOTES
   C library/kernel differences
       The GNU C library fchmodat() wrapper function implements the POSIX-
       specified interface described in this page.  This interface differs
       from the underlying Linux system call, which does not have a flags
       argument.

   Glibc notes
       On older kernels where fchmodat() is unavailable, the glibc wrapper
       function falls back to the use of chmod().  When pathname is a relative
       pathname, glibc constructs a pathname based on the symbolic link in
       /proc/self/fd that corresponds to the dirfd argument.
```

slackner:
I've dropped the support for `flags == 0` and added a comment similar to the other syscallcompat wrappers.

The glibc trick might be similar, but I haven't seen used this trick in combination with `O_PATH` yet. The manpage for `open` also explicitly writes that `fchmod` doesn't work, so I was a bit surprised to find that `chmod` on `/proc/self/fd/...` does.

```
       O_PATH (since Linux 2.6.39)
              Obtain a file descriptor that can be used for two purposes: to
              indicate a location in the filesystem tree and to perform
              operations that act purely at the file descriptor level.  The
              file itself is not opened, and other file operations (e.g.,
              read(2), write(2), fchmod(2), fchown(2), fgetxattr(2),
              ioctl(2), mmap(2)) fail with the error EBADF.
```

rfjakob:
By the way, why don't we use

/proc/self/fd/$dirfd/$path

?

slackner:
That would not work. Then `Chmod()` might resolve `path`, if it points to a symlink.

rfjakob:
Lchmod()

slackner:
But unfortunately that doesn't exist on Linux. You will only find BSD manpages.

rfjakob:
Ah, I see.

rfjakob:
I followed the fchmodat2 discussion at https://patchwork.kernel.org/patch/9596301/ , which ultimately led me to how qemu solves the problem: https://git.qemu.org/?p=qemu.git;a=blob;f=hw/9pfs/9p-local.c;h=08e673a79c2299d535e5cee4317af186ace80f6f;hb=HEAD#l371

I think we need an extra fstat before the chmod.

slackner:
I've now added an explicit test for symlinks, which is definitely a bit safer. Based on my tests, however, it would have already worked before. `Chmod` returned an error (instead of following the symlink target) - at least, if I didn't make any other mistake during my tests.

rfjakob:
I added a Fchmodat  symlink test 8253c55386e59295ea48dfc1b9beea4d785fcbbd . I found something pretty interesting: When you comment-out the `if a.IsSymlink()` check, the permissions on the symlink itself get changed!
```
/tmp/gocryptfs-test-parent/syscallcompat094547068$ ls -l | grep TestFchmodat_Symlink
l-wx-wx-wx. 1 jakob jakob   20 Jan  7 23:05 TestFchmodat_Symlink -> TestFchmodat_Regular
```
I don't think I have ever seen this.

rfjakob:
Your change merged as https://github.com/rfjakob/gocryptfs/commit/4170ef00f32b3943a75f1c85c2b21dbe27ba30cd , thanks!

slackner:
That sounds indeed a bit weird. Thanks for adding the tests!

BTW: When do you plan to release the next -rc? Two things that would be nice to get in before, if we still have sufficient time to address them:
* #327 ("Possible race-conditions between file creation and Fchownat")
* Verify, that we didn't break anything on macOS - did you already look into the Travis CI idea (mentioned in #307)

rfjakob:
By the way, do you mean with

> Note that we can neither use Fchmodat (since fd is not necessarily a directory)

? Why would it help if fd is a directory?

slackner:
For directories, we could also use `Fchmodat` with `path="."`, but not sure if that is necessarily better.

rfjakob:
So I'm back at my day job this week, next rc on the weekend I guess. I absolutely want to fix #327 as it is a security issue.

PR #335, opened 2019-01-09
More fixes for -allow_other

slackner:
Fixes the remaining issues of https://github.com/rfjakob/gocryptfs/issues/327.

Please consider these as draft patches (and adapt them if necessary) - some of the changes (e.g., removing `Chown` for `gocryptfs.iv` files) might have side effects I am not aware of.

### DIFF
--- a/internal/fusefrontend/fs_dir.go
+++ b/internal/fusefrontend/fs_dir.go
@@ -64,6 +64,12 @@ func (fs *FS) Mkdir(newPath string, mode uint32, context *fuse.Context) (code fu
 		return fuse.ToStatus(err)
 	}
 	defer syscall.Close(dirfd)
+	// Don't set the full mode before we have set the correct owner.
+	// See Create FUSE call for more details.
+	origMode := mode
+	if fs.args.PreserveOwner {
+		mode = 0000
+	}
 	if fs.args.PlaintextNames {
 		err = syscallcompat.Mkdirat(dirfd, cName, mode)
 		if err != nil {
@@ -74,14 +80,23 @@ func (fs *FS) Mkdir(newPath string, mode uint32, context *fuse.Context) (code fu
 			err = syscallcompat.Fchownat(dirfd, cName, int(context.Owner.Uid),
 				int(context.Owner.Gid), unix.AT_SYMLINK_NOFOLLOW)
 			if err != nil {
-				tlog.Warn.Printf("Mkdir: Fchownat failed: %v", err)
+				tlog.Warn.Printf("Mkdir %q: Fchownat %d:%d failed: %v", cName, context.Owner.Uid, context.Owner.Gid, err)
+				// In case of a failure, we don't want to proceed setting more
+				// permissive modes.
+				return fuse.ToStatus(err)
+			}
+		}
+		// Set mode
+		if origMode != mode {
+			err = syscallcompat.Fchmodat(dirfd, cName, origMode, unix.AT_SYMLINK_NOFOLLOW)
+			if err != nil {
+				tlog.Warn.Printf("Mkdir %q: Fchmodat %#o -> %#o failed: %v", cName, mode, origMode, err)
 			}
 		}
 		return fuse.OK
 	}
 
 	// We need write and execute permissions to create gocryptfs.diriv
-	origMode := mode
 	mode = mode | 0300
 
 	// Handle long file name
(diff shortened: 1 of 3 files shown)


internal/nametransform/longnames.go
```diff
@@ -129,7 +129,7 @@ func (n *NameTransform) WriteLongNameAt(dirfd int, hashName string, plainName st
 
 	// Write the encrypted name into hashName.name
 	fdRaw, err := syscallcompat.Openat(dirfd, hashName+LongNameSuffix,
-		syscall.O_WRONLY|syscall.O_CREAT|syscall.O_EXCL, 0600)
+		syscall.O_WRONLY|syscall.O_CREAT|syscall.O_EXCL, 0400)
```
rfjakob: Merged

internal/fusefrontend/fs_dir.go
```diff
@@ -104,24 +119,27 @@ func (fs *FS) Mkdir(newPath string, mode uint32, context *fuse.Context) (code fu
...
 		if err != nil {
-			tlog.Warn.Printf("Mkdir: Fchownat 2 failed: %v", err)
+			tlog.Warn.Printf("Mkdir %q: Fchownat(2) %d:%d failed: %v", cName, context.Owner.Uid, context.Owner.Gid, err)
+		}
+	}
+	// Set mode
+	if origMode != mode {
+		err = syscallcompat.Fchmodat(dirfd, cName, origMode, unix.AT_SYMLINK_NOFOLLOW)
+		if err != nil {
+			tlog.Warn.Printf("Mkdir %q: Fchmodat %#o -> %#o failed: %v", cName, mode, origMode, err)
```
rfjakob: Merged

internal/fusefrontend/fs_dir.go
```diff
@@ -124,16 +123,11 @@ func (fs *FS) Mkdir(newPath string, mode uint32, context *fuse.Context) (code fu
...
+			tlog.Warn.Printf("Mkdir %q: Fchownat %d:%d failed: %v", cName, context.Owner.Uid, context.Owner.Gid, err)
 			// In case of a failure, we don't want to proceed setting more
 			// permissive modes.
 			return fuse.ToStatus(err)
 		}
-		err = syscallcompat.Fchownat(dirfd, filepath.Join(cName, nametransform.DirIVFilename),
-			int(context.Owner.Uid), int(context.Owner.Gid), unix.AT_SYMLINK_NOFOLLOW)
-		if err != nil {
-			tlog.Warn.Printf("Mkdir %q: Fchownat(2) %d:%d failed: %v", cName, context.Owner.Uid, context.Owner.Gid, err)
-		}
```
rfjakob: Not sure about the side effects either. I'll run xfstests against it, and it if xfstests is happy, I'm happy.

slackner:
> Looking at 
 https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/vfs.txt?h=v5.0-rc1#n433 , I don't think this can happen:
>
> mknod: called by the mknod(2) system call to create a device (char,
> block) inode or a named pipe (FIFO) or socket.

According to the manpage of `mknod`, the file type can be `S_IFREG`. See e.g., https://www.systutorials.com/docs/linux/man/2-mknod/:

> The file type must be one of S_IFREG, S_IFCHR, S_IFBLK, S_IFIFO or S_IFSOCK to specify a regular file (which will be created empty), character special file, [...]

Even if the kernel internally maps the creation of `S_IFREG` inodes to the `Create` FUSE call (we can adjust the commit message and comment if that is the case - but we should be sure that it happens on all systems!), it still seems a bit dangerous. Doesn't it still suffer from the same problem that other users might be able to open a handle? Assume, e.g., that a user creates a FIFO with `0404` permissions, i.e., noone in his group should be able to access it. As long as it is still owned by `root:root`, other users are able to open it.

rfjakob:
> Assume, e.g., that a user creates a FIFO with 0404 permissions, i.e., noone in his group should be able to access it. As long as it is still owned by root:root, other users are able to open it.

Good point. I will adjust the commit message and merge.

rfjakob:
Hmm, something is wrong with fcaca5fc94d981aa637beb752edc8cb3c2265e96 , it fails in xfstests generic/314 :

```
generic/314 6s ... - output mismatch (see /home/jakob.donotbackup/code/fuse-xfstests/results//generic/314.out.bad)
    --- tests/generic/314.out	2018-06-27 21:12:13.650235083 +0200
    +++ /home/jakob.donotbackup/code/fuse-xfstests/results//generic/314.out.bad	2019-01-09 22:46:43.474164276 +0100
    @@ -1,2 +1,2 @@
     QA output created by 314
    -drwxr-sr-x subdir
    +drwxr-xr-x subdir
```

rfjakob:
Fixed by https://github.com/rfjakob/gocryptfs/pull/340

Issue #339, opened 2019-01-12
PreserveOwner: switch to Setreuid/Setregid?


rfjakob:
This ticket tries tries to summarize the discussion about whether we should use Setreuid/Setregid for handling PreserveOwner correctly. (PreserveOwner = running as root with `-allow_other`) 

Getting all corner-cases right with allow_other is tricky, by calling Setreuid+Setregid we let the kernel handle all the complexity. Open bugs: #327 #336 #337 #338

On Linux, Setreuid+Setregid only affects one thread, so this would work:
1) runtime.LockOSThread
2) Setregid
3) Setreuid
4) runtime.UnlockOSThread

MacOS has settid: https://github.com/rfjakob/gocryptfs/issues/327#issuecomment-452359224

> On macOS there is apparently settid, but we would have to write our own syscall wrappers. Some resources that might help (if we decide to go into this direction):

I'm ok with dropping support for PreserveOwner on MacOS and see if anybody complains.

Proposed patches for Create() by @slackner 

Approach 1: call Setreuid in Create(): http://ix.io/1xX9

> this would be a fix for Create (fixing the mode race condition + fixing all the SGID problems)

Approach 2: add OpenatUser helper: http://ix.io/1xY2

> The approach in my previous draft patch has the disadvantage that it also affects *.name files. While this might seem nice at first (it would be consistent with reverse mode, where *.name files and gocryptfs.diriv files also inherit the uid/gid of the main file/directory), it has the disadvantage of making other operations like Chown, Rename, ... much more complicated. Whenever WriteLongNameAt is used we have to switch to the correct user, even if it was not necessary before.
> 
> A slightly different approach based on changing effective permissions is shown in this patch: http://ix.io/1xY2. By adding new syscallcompat wrappers, we can put all the ugly code in a separate file and minimize the region during which we have to lock OSThread. Similar syscall wrappers would be necessary for other functions that depend on the uid/gid.
> 
> What are your thoughts? Does using effective permissions make things sufficiently easier, or should we stick to the previous Fchown approach (and handle SGID inheritance manually)?

rfjakob:
Performance-testing with this mockup shows no change:

```diff
diff --git a/internal/syscallcompat/sys_linux.go b/internal/syscallcompat/sys_linux.go
index 5c180ec..7a8d032 100644
--- a/internal/syscallcompat/sys_linux.go
+++ b/internal/syscallcompat/sys_linux.go
[...]

PR #340, opened 2019-01-12
Switch to Setreuid/Setregid method instead of manually changing ownership of files.

slackner:
Draft patch for the `Sereuid`/`Setregid` method on Linux. On Darwin, for now, just ignore the ownership problem at all. This will have to be adressed before `-allow_other` is usable again.

See https://github.com/rfjakob/gocryptfs/issues/339

### DIFF
--- a/internal/fusefrontend/fs.go
+++ b/internal/fusefrontend/fs.go
@@ -192,7 +192,7 @@ func (fs *FS) openWriteOnlyFile(dirfd int, cName string, newFlags int) (*File, f
 		return nil, fuse.ToStatus(err)
 	}
 	// The cast to uint32 fixes a build failure on Darwin, where st.Mode is uint16.
-	perms := uint32(st.Mode & 0777)
+	perms := uint32(st.Mode)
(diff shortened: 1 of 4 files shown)


slackner:
Note that I think hardcoding UID 0 and GID 0 is not really a problem (there was a FIXME before), since it is also hardcoded in other files.

rfjakob:
I want to merge this. Code looks good and I like the OpenAtUser idea very much. xfstests are running NOW.

rfjakob:
Results: https://gist.github.com/rfjakob/54897967968e34cad427366da0e22bd4

slackner:
Are any of these failures new? I'm wondering a bit about the `Low memory!` and `fallocate: Software caused connection abort` warnings.

rfjakob:
The low memory thing with generic/213 happens every now and then, the test causes gocryptfs to use 4GB RSS. I checked the behavoir before this PR, and it is the same. As I have 8GB of RAM in the box, this OOMs when I have too many browser tabs open, and/or too much stuff on tmpfs.

PR #341, opened 2019-01-12
fusefrontend: Do not pass unsupported flags to Faccessat on macOS.

slackner:
Fixes mounting of forward mounts on macOS High Sierra.

(Feel free to adjust the patch if you prefer a different version, e.g., filtering it in a syscall wrapper.)


slackner:
I just noticed that there is a better way to fix it, patch updated.

rfjakob:
Merged as d44fe89ba4f3252c5bd00c4f7730197732f2a26a

PR #342, opened 2019-01-13
syscallcompat: Use the same *User() syscall wrappers on both Linux and macOS.

slackner:
I'm surprised a bit that exactly the same method we use on Linux also works on macOS.

Should we move it to a separate file?

### DIFF
--- a/internal/syscallcompat/sys_common.go
+++ b/internal/syscallcompat/sys_common.go
@@ -2,9 +2,12 @@ package syscallcompat
 
 import (
 	"bytes"
+	"runtime"
 	"syscall"
 
 	"golang.org/x/sys/unix"
+
+	"github.com/hanwen/go-fuse/fuse"
 )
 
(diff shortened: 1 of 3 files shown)


rfjakob:
Is it documented somewhere that Setregid only affects one thread on MacOS?

slackner:
Thats a good point. Please wait before merging this patch.

slackner:
Maybe we should even add a test for that? We could also run into trouble if at some point in the future, the go syscalls change their behavior and try to be more clever.

rfjakob:
The [man page](https://opensource.apple.com/source/xnu/xnu-4903.221.2/bsd/man/man2/setreuid.2) says:

> The real and effective user IDs of the current **process** are set according to the arguments.

slackner:
Hm, okay, then I really have to implement `settid`. Will open a new pull request when this is done.

Issue #350, opened 2019-01-15
MacOS: test failure in TestUtimesNano


slackner:
```
--- FAIL: TestUtimesNano (0.00s)
    matrix_test.go:628: "/var/folders/6x/tlqzyq0j38j9xw6qkl8p3jcm0000gq/T/gocryptfs-test-parent/720506002/default-plain/utimesnano": no such file or directory
FAIL
FAIL	github.com/rfjakob/gocryptfs/tests/matrix	3.027s
```

The exact reasons for this issue are not known yet. Some initial analysis was done in https://github.com/rfjakob/gocryptfs/pull/343:

> Hm, this bug is really mysterious. Here is what I found out so far:
> 
> * Yes, it only worked because of defer. When closing the file immediately the test fails. This can be explained based on the go-fuse code, which uses file-based operations as long as at least one FD is open.
> * Disabling the fuse cache (setting all timeouts to 0) does not help.
> * Disabling the directory fd cache does not help.
> * Adding a time.Sleep of 2 seconds after creating the file also doesn't help.
>
> However, to make sure that the Utimens call is forwarded at all, I put debug messages into the FUSE function. They confirm that ENOENT is returned from the unix.UtimesNanoAt call?!

slackner:
Output with `-fusedebug` and `-d`:
https://gist.githubusercontent.com/slackner/bbeabbc4aeb4b85b103a98ba048ec0a8/raw/18e13c2689aef0f39b84fbcb30c17349fe259341/gistfile1.txt

slackner:
Apparently Utimens is internally a `Setattr` call (mapped in `fuse/nodefs/fsops.go`, function `SetAttr`), however, that still doesn't explain why it doesn't show up.

rfjakob:
Hmm, and syscall.UtimesNano calls setattrlistTimes on MacOS: [code](https://github.com/golang/go/blob/55715c1738adc0a9b6a7037390a700763365e284/src/syscall/syscall_bsd.go#L507)

Maybe setattrlistTimes just does not work at all on FUSE on MacOS?

slackner:
The link above points to `UtimesNano`, but we are using `UtimesNanoAt`. Source code is here:
https://github.com/golang/sys/blob/1b2967e3c290b7c545b3db0deeda16e9be4f98a2/unix/syscall_bsd.go#L562

```
func UtimesNanoAt(dirfd int, path string, ts []Timespec, flags int) error {
	if ts == nil {
[...]

PR #352, opened 2019-01-16
Rework the Utimens handling on macOS.

slackner:
For Linux, everything effectively stays the same. For both path-based and
fd-based Utimens() calls, we use unix.UtimesNanoAt(). To avoid introducing
a separate syscall wrapper for futimens() (as done in go-fuse, for example),
we instead use the /proc/self/fd - trick.

On macOS, this changes quite a lot:

* Path-based Utimens() calls were previously completely broken, since
  unix.UtimensNanoAt() ignores the passed file descriptor. Note that this
  cannot be fixed easily since there IS no appropriate syscall available on
  macOS prior to High Sierra (10.13). We emulate this case by using
  Fchdir() + setattrlist().

* Fd-based Utimens() calls were previously translated to f.GetAttr() (to
  fill any empty parameters) and syscall.Futimes(), which does not does
  support nanosecond precision. Both issues can be fixed by switching to
  fsetattrlist().

Fixes https://github.com/rfjakob/gocryptfs/issues/350

---------------

The first commit is cleanup and could also be omitted, if there are any objections. The second commit should be considered a draft patch, however, it fixes all the `utimens` related errors on macOS.

### DIFF
--- a/internal/syscallcompat/sys_darwin.go
+++ b/internal/syscallcompat/sys_darwin.go
@@ -2,8 +2,11 @@ package syscallcompat
 
 import (
 	"log"
+	"path/filepath"
 	"runtime"
 	"syscall"
+	"time"
+	"unsafe"
 
 	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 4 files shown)


internal/syscallcompat/sys_darwin.go
```diff
@@ -125,6 +146,72 @@ func MkdiratUser(dirfd int, path string, mode uint32, context *fuse.Context) (er
...
+		unsafe.Sizeof(attributes), 0)
+}
+
+// UtimesNanoAtNofollow is like UtimesNanoAt but never follows symlinks.
+//
+// Unfortunately we cannot use unix.UtimesNanoAt since it is broken and just
+// ignores the provided 'dirfd'. In addition, it also lacks handling of 'nil'
+// pointers (used to preserve one of both timestamps).
+func UtimesNanoAtNofollow(dirfd int, path string, a *time.Time, m *time.Time) (err error) {
+	if !filepath.IsAbs(path) {
```
slackner: Note that I didn't put it in `emulation.go` since `setattrlist` does not exist on Linux. We might want to rework the directory structure anyway, now that it only contains a single function.

rfjakob:
Comment added and merged

PR #401, opened 2019-05-21
Add missing defer close file

leilaes:
This PR fixes missing defer close files [as specified in Effective Go](https://golang.org/doc/effective_go.html#defer). These issues were found by [this tenet](https://github.com/codelingo/codelingo/blob/master/tenets/codelingo/effective-go/defer-close-file/codelingo.yaml). Your [codelingo.yaml file](https://github.com/leilaes/gocryptfs/blob/master/codelingo.yaml) has been updated as this tenet is part of our effective-go bundle now :) 
Please note that the following cases were also found by the tenet, but have not been changed in this PR as the files have been opened in a loop:
https://github.com/leilaes/gocryptfs/blob/master/tests/matrix/concurrency_test.go#L107
https://github.com/leilaes/gocryptfs/blob/master/tests/defaults/diriv_test.go#L48

### DIFF
--- a/codelingo.yaml
+++ b/codelingo.yaml
@@ -1,4 +1,3 @@
 tenets:
 - import: codelingo/effective-go
 - import: codelingo/code-review-comments
-- import: codelingo/rfjakob-gocryptfs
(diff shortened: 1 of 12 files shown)


rfjakob:
Hi! Not sure about this. The biggest problem is that it removes error handling in several cases. file.Close() can return an error!

Then, in some cases you actually want to close the file **now** - keeping it open keeps the file from being deleted.

In short, I don't think it makes sense to replace direct file.Close() with a deferred version.

PR #408, opened 2019-07-24
Prevent local user ID from appearing in source tarballs

dhirsbrunner:
The local user ID (1026 jakob) appears in the source tarballs gocryptfs_v1.7_src.tar.gz and gocryptfs_v1.7_src-deps.tar.gz as the owner of VERSION, Documentation, and vendor. This issue is already fixed for the binary releases by commit 07f57314afb260d6b14227b932d66345c55ffab3, and the solution here is the same: use "tar --owner=root --group=root".


rfjakob:
Goid catch, thanks!

PR #419, opened 2019-09-06
Provide a manpage for statfs tool.

lechner:
From Debian. The `render` command in `MANPAGE-render.bash` is untested. We generate the man pages directly with `pandoc`.


rfjakob:
Wow, I did not expect this to be useful to somebody! OTOH, the reason it exists is that there seemed to be no user space tool to get the statfs information.

rfjakob:
Merged and expanded a little, thanks! https://github.com/rfjakob/gocryptfs/blob/master/Documentation/MANPAGE-STATFS.md

PR #420, opened 2019-09-06
Fix 'cannot adjust line' warning from man in gocryptfs manpage.

lechner:
When generating `man` pages, `pandoc` marks indented code blocks with the
`roff` macro `.nf`. That avoids a warning from `man` related to the long
line about the master key.

### DIFF
--- a/Documentation/MANPAGE.md
+++ b/Documentation/MANPAGE.md
@@ -224,9 +224,10 @@ settings have to be passed on the command line: `-aessiv` when you
 mount a filesystem that was created using reverse mode, or
 `-plaintextnames` for a filesystem that was created with that option.
 
-Examples:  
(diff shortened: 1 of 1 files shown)


rfjakob:
I can repro the warning like this (if the terminal window is narrow!):
```
gocryptfs/Documentation$ man ./gocryptfs.1 > /dev/null
<standard input>:282: warning [p 4, 8.8i]: cannot adjust line
<standard input>:282: warning [p 4, 9.0i]: can't break line
```

rfjakob:
Patch gets rid of the warning, output looks better too: Before:
```
       Examples:
       -mas‐
       terkey=6f717d8b-6b5f8e8a-fd0aa206-778ec093-62c5669b-abd229cd-241e00cd-b4d6713d
       -masterkey=stdin
```

After:
```
       Examples:

              -masterkey=6f717d8b-6b5f8e8a-fd0aa206-778ec093-62c5669b-abd229cd-241e00cd-b4d6713d
              -masterkey=stdin
```

PR #425, opened 2019-09-14
Show undecryptable filenames if they match supplied regex

### LINKED ISSUE #393: Show undecryptable file names in the dir listing
Dropbox etc handle synchronisation conflicts by preserving both versions and appending something like `(conflict)` to the file name. 

The encrypted file will look like `9sv5JFmMzhvBE7faKVZZdQ (conflict)`. This does not decrypt properly, and gocryptfs will log an error and ignore the file. So basically, those `(conflict)` files become invisible.

We should find a way to make these files visible to the user. Simplest option is listing them as something like
```
GOCRYPTFS_INVALID_NAME_1234 9sv5JFmMzhvBE7faKVZZdQ (conflict)
```

UnicodingUnicorn:
Should resolve #393. I do not get what functionalities require tests written for them in this codebase, so I have neglected that for now. Please advise.


slackner:
I'm not sure how that is supposed to work since `BypassRegex` is never set anywhere.

UnicodingUnicorn:
Sorry, my bad. Forgot to add those files when committing.

internal/nametransform/names.go
```diff
@@ -53,6 +56,10 @@ func New(e *eme.EMECipher, longNames bool, raw64 bool) *NameTransform {
 // DecryptName decrypts a base64-encoded encrypted filename "cipherName" using the
 // initialization vector "iv".
 func (n *NameTransform) DecryptName(cipherName string, iv []byte) (string, error) {
+  if n.BypassRegex != nil && n.BypassRegex.MatchString(cipherName) {
```
rfjakob: This should only trigger if decrypting the name fails. In other words, try to decrypt first, then check if Bypass matches. I suggest to create a new helper function, `(n *NameTransform) decryptName(cipherName string, iv []byte)` that tries to do the regular decryption. `DecryptName` then only calls `decryptName` and check Bypass when `decryptName` fails.

UnicodingUnicorn:
@rfjakob may I request another review please?

UnicodingUnicorn:
@rfjakob I confess I am a bit stuck doing the test. I have pushed what I have for now, with the test called "TestBypass" in `tests/cli/cli_test.go`.

I have tested the code manually and the bypass seems to work. However, when I try to do the automated test, the rejected name (INVALID_GOCRYPTFS_NAME ...) is lost. I have been able to trace it up to `OpenDir` in `internal/fusefrontend/fs_dir.go`, but have been unable to trace it further since I have no idea where that is called.

Is `ioutil.ReadDir` the right way to do things? I don't think I have that good an idea how it pretends to be a filesystem too.

rfjakob:
Looked at the code, and you are almost there. Don't use `ioutil.ReadDir`, use this instead:
```
	f, err := os.Open(mnt)
	if err != nil {
		t.Fatal(err)
	}
	defer f.Close()
	names, err := f.Readdirnames(0)
```
The problem is that `ioutil.ReadDir` will try to get file size etc, this fails for the `INVALID_GOCRYPTFS_NAME` file, and ReadDir will suppress the entry.

rfjakob:
Complete test function could look like this (tested and works):

```go
func TestBypass(t *testing.T) {
	dir := test_helpers.InitFS(t)
	mnt := dir + ".mnt"
	test_helpers.MountOrFatal(t, dir, mnt, "-namedecr=*", "-extpass=echo test")
	defer test_helpers.UnmountPanic(mnt)

	file := mnt + "/normalEncryptedFile"
	err := ioutil.WriteFile(file, []byte("somecontent"), 0600)
	if err != nil {
		t.Fatal(err)
	}
	const invalidName = "plaintextFileInCiphertextDir"
	file1 := dir + "/" + invalidName
	err = ioutil.WriteFile(file1, []byte("somecontent"), 0600)
	if err != nil {
		t.Fatal(err)
	}
	// Check that plaintextFileInCiphertextDir shows up in directory listing
	f, err := os.Open(mnt)
	if err != nil {
		t.Fatal(err)
	}
	defer f.Close()
	names, err := f.Readdirnames(0)
	found := ""
	for _, v := range names {
		if strings.Contains(v, invalidName) {
			found = v
		}
	}
	if found == "" {
		t.Errorf("did not find %s in %v", invalidName, names)
	}
	t.Log(names)
}
```

rfjakob:
gocryptfs$ go test -run TestBypass -v ./tests/cli
=== RUN   TestBypass
2020/01/25 16:40:14 Unimplemented opcode POLL
--- PASS: TestBypass (0.03s)
    cli_test.go:649: [INVALID_GOCRYPTFS_NAME plaintextFileInCiphertextDir normalEncryptedFile]
PASS
ok  	github.com/rfjakob/gocryptfs/tests/cli	0.038s

rfjakob:
Merged as https://github.com/rfjakob/gocryptfs/commit/9ec042f2f62bc95154d6c8b3215a2e7853f8f5c6 with some small adjustments, thanks!

rfjakob:
Go for the bounty at https://www.bountysource.com/issues/72827980-show-undecryptable-file-names-in-the-dir-listing !

PR #432, opened 2019-10-24
Logging corrections...

madscientist42:
Logging, since we're on Linux/OSX/*BSD, should just be
a matter of switching the writer for the logging in tlog
and the default logger to the syslog one.  We need to,
however, do a .Dial call because .New appears to not
do what one intends- it doesn't establish a connection
to the syslog daemon unless you already have one from
.Dial

So, with that, we cleaned out the piped shell call to
logger, which adds to the overhead, isn't needed and
open up the UNIX domain socket hook to syslog with
log/syslog instead.  Seems to work for the cases
I needed for things.  Future iterations can accept
an option for -tcp/-udp and an IP address for
a network logging option if needed now.

### DIFF
--- a/daemonize.go
+++ b/daemonize.go
@@ -62,7 +62,11 @@ func forkChild() int {
 }
 
 // redirectStdFds redirects stderr and stdout to syslog; stdin to /dev/null
-func redirectStdFds() {
+//
+// (We're passing in a tagname (basically the mountpoint...) to be clearer
+//  than a PID which will not be the same from boot to boot and making it
+//  be consistent with the normal logging... )
+func redirectStdFds(tagName string) {
 	// Create a pipe pair "pw" -> "pr" and start logger reading from "pr".
 	// We do it ourselves instead of using StdinPipe() because we need access
 	// to the fd numbers.
@@ -71,7 +75,7 @@ func redirectStdFds() {
 		tlog.Warn.Printf("redirectStdFds: could not create pipe: %v\n", err)
 		return
 	}
-	tag := fmt.Sprintf("gocryptfs-%d-logger", os.Getpid())
+	tag := fmt.Sprintf("gocryptfs[%s]", tagName)
 	cmd := exec.Command("logger", "-t", tag)
 	cmd.Stdin = pr
 	err = cmd.Start()
--- a/internal/tlog/log.go
+++ b/internal/tlog/log.go
@@ -136,11 +136,15 @@ func init() {
 }
 
 // SwitchToSyslog redirects the output of this logger to syslog.
-func (l *toggledLogger) SwitchToSyslog(p syslog.Priority) {
-	w, err := syslog.New(p, ProgramName)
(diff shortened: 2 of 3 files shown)


rfjakob:
Hi! You say that syslog.New() does not work, that means that logging to syslog is completely broken at the moment?

About the logger, this actually is needed, this is the only way to also capture panics.

madscientist42:
Hm.  It was not logging to syslog directly.  When you don't Dial in, it just acts like a normal logger entity in Go for defensive purposes.  So your catch for the panics there was doing all the logging as best as I could tell.  So, the question then is if there's a way to catch the panics better or if there's no way and we could simplify this.

Annoying...  X-D

Let me re-think the PR here a smidge to see what needs to be properly proposed to you here.

rfjakob:
For the stuff the external logger catches, you should see a "gocryptfs-logger-1234" prefix. Direct syslog should read just "gocryptfs-1234"

madscientist42:
I think what's in the PR right now should reflect what I'd intended to ask you to accept to get the logging to work right, in light of the issue with panics (It's been a little bit since I did any Go coding, I'd forgotten about THAT one...  X-D).

The desire is to have it log properly to syslog under most conditions.   Since we have to use logger to catch panics that're not resolved, I'm wondering if it wouldn't be cleaner to just leave it alone, let it use the to-console logging mode, just strip the colors and the timestamps and let logger catch everything and log that way?

madscientist42:
The change to using the mountpoint is to help make debugging over long-term uses or reboots easier.

A PID is only useful in the scope of a single boot session or a given mount.  It requires people to do sleuthing to figure out what the log is telling them.  Using a mountpoint as specified lets the user clearly see what instance/mount had the problem, etc.

madscientist42:
As for the way you're doing it there, it was never doing the syslog prefixing, only the logger one.  I made them one in the same so that  you have consistent logging.  The other tells you that the logger caught it (Which is my tip that it wasn't doing QUITE what you'd intended, so there is that...) but it adds noise to the log which ultimately only means something if you had the condition I caught out with it.

rfjakob:
Hi, couple of things:

1) Can you provide an example of log line in old log format, and one in new format?

2) Having the `logger` process is undesirable because a lot of metadata is lost (check `journalctl -o json` to see what metadata is collected). For panics it is unavoidable, but not for normal logging.

3) Using the mountpoint path in the tag is OK

rfjakob:
4) The commit message does not seem to match what the patch does anymore

madscientist42:
Sorry about being non-responsive...  Got tied up with a Beta release.  Give
me a bit longer and I can respond, etc. to what I observed, etc.

rfjakob:
Closing for now... Reopen at will

madscientist42:
Will do.  Got VERY sidetracked at work all this time with several fire
drills.  I think I've a better proposal I'll submit with the ticket as a PR
shortly.  :D

PR #433, opened 2019-10-31
add some security by checking permissions on passfile

jerseyd1234:
Add a bit of security by forcing the user to only have read/write permission on any file specified for passfile.  It is a small improvement as it forces the user to keep the file from read by anyone else.


rfjakob:
Not sure about this. The passfile may reside on a FAT-formatted USB stick where permission are not supported at all. I think this adds more pain than gain.

Thanks for the suggestion anyway!

PR #440, opened 2019-12-09
Starting with fuse 3.0 the nonempty option has been removed.

juddmon:
I ran into this problem in Ubuntu 20.04 beta.

The code for libfuse >= 3.0 had the nonempty mount option removed from fusermount because the standard mount option does not have this option.  This caused gocryptfs to fail when this option is used (fuse doesn't allow that option).  Not using it will not work either if the mountpoint is not empty because the check "isEmptyDir" failed.

http://fuse.996288.n3.nabble.com/ANNOUNCE-libfuse-3-0-0-has-been-released-td14403.html

### DIFF
--- a/cli_args.go
+++ b/cli_args.go
@@ -157,7 +157,7 @@ func parseCliOpts() (args argContainer) {
 		"Only works if user_allow_other is set in /etc/fuse.conf.")
 	flagSet.BoolVar(&args.reverse, "reverse", false, "Reverse mode")
 	flagSet.BoolVar(&args.aessiv, "aessiv", false, "AES-SIV encryption")
-	flagSet.BoolVar(&args.nonempty, "nonempty", false, "Allow mounting over non-empty directories")
+	flagSet.BoolVar(&args.nonempty, "nonempty", false, "Allow mounting over non-empty directories (deprecated for fuse >=3.0)")
 	flagSet.BoolVar(&args.raw64, "raw64", true, "Use unpadded base64 for file names")
 	flagSet.BoolVar(&args.noprealloc, "noprealloc", false, "Disable preallocation before writing")
 	flagSet.BoolVar(&args.speed, "speed", false, "Run crypto speed test")
--- a/mount.go
+++ b/mount.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"log"
@@ -12,6 +13,8 @@ import (
 	"os/signal"
 	"path"
 	"path/filepath"
+	"regexp"
+	"strconv"
 	"runtime"
 	"runtime/debug"
 	"strings"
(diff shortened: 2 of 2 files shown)


rfjakob:
Hi! Thanks for the patch! The version parsing logic looks a little scary to me, I tried to have a simpler solution. Can you try https://github.com/rfjakob/gocryptfs/commit/6eee5f24655e03546e949b5559a5392861be78c1 and see if it fixes your problem?

Also, gocryptfs will keep the `-nonempty` option, as I think it is useful. Mounting over non-empty directories should still be rejected by default, also on libfuse3.

juddmon:
> Hi! Thanks for the patch! The version parsing logic looks a little scary to me, I tried to have a simpler solution. Can you try [6eee5f2](https://github.com/rfjakob/gocryptfs/commit/6eee5f24655e03546e949b5559a5392861be78c1) and see if it fixes your problem?
> 
> Also, gocryptfs will keep the `-nonempty` option, as I think it is useful. Mounting over non-empty directories should still be rejected by default, also on libfuse3.

Mounting over non-empty directories **is allowed** by default in libfuse 3.0 to match the behavior of mount.

There is a changelog entry for libfuse3.0 that explains why the -nonempty option breaks with libfuse3.
http://fuse.996288.n3.nabble.com/ANNOUNCE-libfuse-3-0-0-has-been-released-td14403.html

> The `-o nonempty` mount point has been removed, mounting over
  non-empty directories is now always allowed. This brings the
  behavior of FUSE file systems in-line with the behavior of the
  regular `mount` command.

> File systems that do not want to allow mounting to non-empty
  directories should perform this check themselves before handing
  control to libfuse.
Here is the error I get with this commit:
`$ gocryptfs -nonempty forward forward-mount/
Password: 
Decrypting master key
/usr/bin/fusermount: unknown option 'nonempty'
fuse.NewServer failed: fusermount exited with code 256
`
This is on Ubuntu 20.04.  I am happy to keep testing for you until we get this right.  My PR worked, but I do agree that I wasn't real happy with the version checking and was hoping you knew a better way.

juddmon:
I did not copy the binary built from 6eee5f2 to /usr/bin/gocryptfs.  After I did this it works and I am happy with it.  Thanks for accepting part of my patch and help!

rfjakob:
Thanks for testing! Merged to master.

> Mounting over non-empty directories should still be rejected by default, also on libfuse3.

Sorry, that was unclear! gocryptfs will keep rejecting nonempty directories per default, also when using libfuse3. This is checked inside gocryptfs by the `isEmptyDir()` function and does not depend on libfuse behavoir.

juddmon:
> > Mounting over non-empty directories should still be rejected by default, also on libfuse3.
> 
> Sorry, that was unclear! gocryptfs will keep rejecting nonempty directories per default, also when using libfuse3. This is checked inside gocryptfs by the `isEmptyDir()` function and does not depend on libfuse behavoir.
Yea, that wording threw me off for a bit and made me re-explain, but I eventually knew what you meant.  Will I get a little credit in the git log or anywhere?

PR #474, opened 2020-05-11
Badnameregex

DerDonut:
Changed bad name resolving behavior to full regex mode. Is related to #393 . Includes modification of test case.

### DIFF
--- a/internal/nametransform/names.go
+++ b/internal/nametransform/names.go
@@ -5,7 +5,7 @@ import (
 	"bytes"
 	"crypto/aes"
 	"encoding/base64"
-	"path/filepath"
+	"regexp"
 	"syscall"
 
 	"github.com/rfjakob/eme"
@@ -37,7 +37,7 @@ type NameTransform struct {
 	// on the Raw64 feature flag
 	B64 *base64.Encoding
 	// Patterns to bypass decryption
-	BadnamePatterns []string
+	BadnamePatterns []*regexp.Regexp
 }
 
 // New returns a new NameTransform instance.
@@ -59,10 +59,16 @@ func (n *NameTransform) DecryptName(cipherName string, iv []byte) (string, error
 	res, err := n.decryptName(cipherName, iv)
 	if err != nil {
 		for _, pattern := range n.BadnamePatterns {
-			match, err := filepath.Match(pattern, cipherName)
-			if err == nil && match { // Pattern should have been validated already
-				return "GOCRYPTFS_BAD_NAME " + cipherName, nil
-			}
+            pos := pattern.FindStringIndex(cipherName)
+            if pos != nil {
+                // concat cipherName without pattern match
+                concatCipherName := cipherName[0:pos[0]] + cipherName[pos[1]:len(cipherName)]
+                res, err := n.decryptName(concatCipherName, iv)
+                // Add pattern always at end
+                if err == nil { // Pattern should have been validated already
+                    return res + cipherName[pos[0]:pos[1]] + "(GOCRYPTFS_BAD_NAME)", nil
+                }
+            }
 		}
 	}
 	return res, err
--- a/mount.go
+++ b/mount.go
@@ -19,6 +19,7 @@ import (
 	"sync/atomic"
 	"syscall"
 	"time"
+    "regexp"
 
 	"golang.org/x/sys/unix"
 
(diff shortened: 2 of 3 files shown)


DerDonut:
Just discovered errors in cli tests. Will fix.

rfjakob:
Hi, sorry for not replying earlier in #393, but are there really cases where wildcards are not enough (which ones)?

DerDonut:
My goal was to be able to rename conflicting files which were created by [Syncthing](https://docs.syncthing.net/users/faq.html#what-if-there-is-a-conflict) when browsing with a file Manager (Dolphin in my case). Syncthing appends a timestamp to a conflicted file since multiple conflicts are possible. This would look like this:
```
plain dir          | cipher dir
  test_file.txt    |   sq7nHcTyK94IMrRYnKPxxQ
  ?                |   sq7nHcTyK94IMrRYnKPxxQ.sync-conflict-20200303-134512
```
To be able to easily know which file was duplicated, it would be helpful to show the original file name in the plain directory also. To achieve this, my thought was to create a specific regex, which is applied to an invalid name. The match of this regex is cut, the name is decrypted and then the match will be appended in plain mode. Only if the decryption is successful, the file will be visible in plain mode. In this case, the parameter would be `-badname=\\.sync-conflict-\\d{8}-\\d{6}` which would extract ".sync-conflict-20200303-134512" from the file name, showing the following view in plain:
```
plain dir                                                          | cipher dir
  test_file.txt                                                    |   sq7nHcTyK94IMrRYnKPxxQ
  test_file.txt.sync-conflict-20200303-134512(GOCRYPTFS_BAD_NAME)  |   sq7nHcTyK94IMrRYnKPxxQ.sync-conflict-20200303-134512
```
If you now rename the file in the file manager manually, the cipher name has to be overwritten, e.g.:
```
plain dir                                      | cipher dir
  test_file.txt                                |   sq7nHcTyK94IMrRYnKPxxQ
  test_file_0303.txt                           |   ZiWY-lznHBN2vp_YKSko274RkCyQSnI71s5FDoKplgY
```
To archieve this, it is necessary to hand over an expression which exactly matches the added text. A global wildcard like * would match the whole filename including the valid cipher part. Only when extracting the exact appended text, the original file name can be decrypted. This solution would also work if the text is added anywhere in the name. Dropbox uses something like text_file(Donut's conflicted file 2020-03-03).txt, so it preserves the file extension. The expression would be something like `-badname=\\(*.\\sconflicted\\sfile\\s\\d{4}-\\d{2}-\\d{2}\\)` (not tested).
The solution would also lower the possibilities of name injections you mentioned in #393. I appended the conflict flag to the end to maintain sorting order near the origin file.

At the moment, my solution would still not show the conflicted files in the file managers since there seems to be no file behind. As i mentioned in #393, I was only capable to list those files in terminal. I currently don't know why the cli tests are not executed, I am pretty sure it worked back in March, but I had to reinstall my machine in between.

rfjakob:
Thanks for the excellent explanation! Use case makes sense to me.

I think this can be automated a little bit, and for `sq7nHcTyK94IMrRYnKPxxQ.sync-conflict-20200303-134512` it is guaranteed to work, because a dot `.` can never occur in a base64-encoded file name. What you can do is:

1) Try do decode base64
2) If that fails, remove last character
3) Go to 1

In the end you will have decoded `sq7nHcTyK94IMrRYnKPxxQ` to `test_file.txt`, and you have the postfix `.sync-conflict-20200303-134512` that you can display.

rfjakob:
> At the moment, my solution would still not show the conflicted files in the file managers since there seems to be no file behind

That is because GETATTR does not handle `test_file.txt.sync-conflict-20200303-134512(GOCRYPTFS_BAD_NAME)`. It will just return an error, and the file manager will think the file has been deleted.

DerDonut:
> Thanks for the excellent explanation! Use case makes sense to me.
> 
> I think this can be automated a little bit, and for `sq7nHcTyK94IMrRYnKPxxQ.sync-conflict-20200303-134512` it is guaranteed to work, because a dot `.` can never occur in a base64-encoded file name. 

This is correct. This would work with all online services I looked up since they use a dot (Syncthing) or Brackets (oneDrive, Google Drive, owncloud/nextcloud, DropBox) and they all seem to only append things to names. Some services try to keep the extension, but this won't matter here because no encoded filename can have a dot in it. I don't know if it's faster if you search for the first non-Base64URL-Encoded char or trying to decode an removing chars from the end like you proposed.
This seems to be a design decision how to present it to the user it. If I am not missing anything, these are the options:

1. Fully automated: invalid name resolving is always applied
2. Single Flag: invalid name resolving is activated by a boolean flag on mount
3. Expression: invalid name resolving is controlled by an expression (regex or filepath)

If your favorite is 1 or 2, I can edit my branch accordingly or create a new one. I'm just now only struggling initializing my IDE to be able to compile forked GO projects.

rfjakob:
I think per default, invalid names should cause a warning but not be shown in the directory. But if the user passes a `-badname` wildcard, and the invalid name matches, we can try to split it automatically. I'm not sure if this is option (1) or (3)   :)

About how to decode or find the first non-base64-character, I think it does not really matter what is faster. Invalid file names will (hopefully) not occur in the thousands. So I would prefer what makes simpler code.

DerDonut:
> [...], we can try to split it automatically. I'm not sure if this is option (1) or (3) :)

Neither of them :). I think I get it now: Use the regex to determine **IF** the file is displayed but check for non-base64URL characters to determine **WHAT** is cut.

I will change my branch accordingly.

rfjakob:
Yes, except that i think wildcard match is good enough

DerDonut:
Ok. Will also work with all-wildcard * with this approach.

DerDonut:
Replaced by #479

PR #479, opened 2020-05-19
Badnameauto

DerDonut:
Changed invalid file decoding and decryption. Function DecryptName now shortens the filename until the filename is decodable and decryptable. Will work with valid **and** invalid Base64URL delimiter (valid delimiter [0-9a-zA-z_\\-]. If the filename is not decryptable at all, it returns the original cipher name with flag suffix
Changed cli tests to generate decryptable and undecryptable file names with correct encrypted content.
Replacing #474, extends #393


DerDonut:
Removed extra decoding check. Leads to shorter code and faster execution.

rfjakob:
Merged, thanks!

PR #485, opened 2020-05-31
contrib/getdents-debug: fix function call missing argument from 22e3eec

jhilliker:
22e3eec15302eac28c1a2ac3f9af29c2c9e82a3c build was broken with error message:
contrib/getdents-debug/getdents/getdents.go:69: Printf format %v reads arg 2, but call has 1 arg

This commit fixes it by adding the missing argument.

Edit: removed # from err msg so github doesnt link it to an issue #


jhilliker:
Codacy/PR Quality Review: error message:
Error getting analysis status for this pull request.

Re-run build?

PR #503, opened 2020-09-05
Fix wrong doc on exit code

Yajo:
According to real world tests, the correct exit code here is 7, so I'm updating the man page to avoid confusion.

See this session:

```bash
[root@132973203c5d /]# gocryptfs -init root
Invalid cipherdir: directory /root not empty
[root@132973203c5d /]# echo $?
7
[root@132973203c5d /]# gocryptfs --version
gocryptfs v1.8.0-Fedora-1.8.0-1.fc32; go-fuse 1.0.1; 2020-05-21 go1.14.2 linux/amd64
```

### DIFF
--- a/Documentation/MANPAGE.md
+++ b/Documentation/MANPAGE.md
@@ -513,7 +513,7 @@ EXIT CODES
 ==========
 
 0: success  
-6: CIPHERDIR is not an empty directory (on "-init")  
+7: CIPHERDIR is not an empty directory (on "-init")  
 10: MOUNTPOINT is not an empty directory  
 12: password incorrect  
 22: password is empty (on "-init")  


rfjakob:
Hi, thanks for the report! I looked at this, and I think gocryptfs returns the wrong exit code. So I fixed that and kept the docs as-is: https://github.com/rfjakob/gocryptfs/commit/993b19c19c26eac480d88cc67bfe7455e67faada .

rfjakob:
PS: The fix will appear in gocryptfs v2.0, which currently being developed in the [gofuse_v2api branch](https://github.com/rfjakob/gocryptfs/commits/gofuse_v2api)

PR #534, opened 2021-01-02
Remove broken instructions for -trezor from README

qix:
It took me a while to figure out why I couldn't get the Trezor implementation working. The only mention of the system is with instructions the changelog, so this just adds a small update to indicate the support is no longer available.

This also appeared in the Ubuntu man page, but not sure where the source of that is.


rfjakob:
Thanks for the report! Addressed slightly differently as https://github.com/rfjakob/gocryptfs/commit/3ab1bcf1c5aacafe4339564ff0a09c349d3201a0 .

PR #579, opened 2021-06-19
MacOS: Add install instructions for macports

MarcelBochtler:
Add installation instructions for [MacPorts](https://www.macports.org/).
See: https://github.com/macports/macports-ports/blob/master/fuse/gocryptfs/Portfile


rfjakob:
Looks good, merged, thanks!

Issue #108, opened 2017-05-21
Feature Request: Fully-deterministic AES-SIV mode


Nodens-:
Hello, great work and amazing performance from my tests so far!

In cppcryptfs documentation this is mentioned:

> "Note: when you mount a filesystem using AES256-SIV in forward mode, any new encryption is done non-deterministcally (as is the case with gocryptfs)."

Is this really the case and if it is, why is it so? In order to avoid copy pasting the same thing here I'll link to the issue there: https://github.com/bailey27/cppcryptfs/issues/15

rfjakob:
Hi there, thanks for the report. A few things:

1) Are you aware that reverse mode is read only? I didn't understand your
synchronisation scheme completely, just wanted to make sure this does not
bite you.

2) Determinism: every 4kb ciphertext block has a 16 byte nonce header.
aes-siv does not strictly need a nonce, but i wanted to keep the file
format the same as for aes-gcm. In aes-siv forward mode, a random value is
stored there, because it improves security somewhat. But I agree that this
should be configurable or even off by default.

Nodens-:
Yes I do understand the read-only part. I am only doing one-way sync of those directories (they contain c++ sources I am actively working on plus a few other things that can't really be encrypted on the local copy). All I want in my scheme is for them to be backed up and encrypted on the remote untrusted location (gdrive) and available to the local network so I can pull copies. Most of them are repositories on my own gitlab server so this serves mostly as an always accessible backup in case something is wrong with the gitlab server and temporary backup of changes that have not been commited to git yet.
 If I need to manually replace a file there as part of some repair operation, it would be nice if the sync program did not detect different files and try to upload again.

Since we're discussing this however would writable reverse be possible at any point? Is there any technical reason on why reverse is ro? Write support would greatly boost what  you can do with syncing etc.

rfjakob:
I think it makes sense to have aes-siv deterministic in forward mode as well. I'll change that in v1.4.

Read-write is technically feasible, it's just not implemented. Having the mount read-only simplifies the code considerably, and for backups you usually *want* read-only access.

rfjakob:
I have started thinking about and working on this. In reverse mode, the file initialisation vector (makes sure that identical files in different directories encrypt differently) is based on the path to the file. Reverse mode is read-only and this works fine.

[...]

PR #592, opened 2021-08-13
Allow deterministic diriv creation via cli flag

Kuroneer:
This PR introduces a new flag (`-deterministicdirivcreation`), which generates the diriv from the first bytes of the sha256 of the plaintext path.

I'm not very familiar with go and I was unable to run the tests (old ubuntu, I tested it by hand compiling it with `docker run -u $(id -u):$(id -g) -v $PWD:/source -w /source -e HOME=/source golang ./build-without-openssl.bash`). That's why I'm initially leaving this as a draft. I hope that the github actions will be able to test it.

### DIFF
--- a/Documentation/MANPAGE.md
+++ b/Documentation/MANPAGE.md
@@ -545,6 +545,11 @@ useful in regression testing.
 
 Applies to: all actions.
 
+#### -zerodiriv
+Create diriv as all-zero files
+
+Applies to: all actions without `-plaintextnames`.
+
 #### \-\-
 Stop option parsing. Helpful when CIPHERDIR may start with a
 dash "-".
--- a/cli_args.go
+++ b/cli_args.go
@@ -30,7 +30,7 @@ type argContainer struct {
 	plaintextnames, quiet, nosyslog, wpanic,
 	longnames, allow_other, reverse, aessiv, nonempty, raw64,
 	noprealloc, speed, hkdf, serialize_reads, forcedecode, hh, info,
-	sharedstorage, devrandom, fsck, one_file_system bool
+	sharedstorage, devrandom, fsck, one_file_system, zerodiriv bool
 	// Mount options with opposites
 	dev, nodev, suid, nosuid, exec, noexec, rw, ro, kernel_cache, acl bool
 	masterkey, mountpoint, cipherdir, cpuprofile,
@@ -179,6 +179,7 @@ func parseCliOpts(osArgs []string) (args argContainer) {
 	flagSet.BoolVar(&args.devrandom, "devrandom", false, "Use /dev/random for generating master key")
 	flagSet.BoolVar(&args.fsck, "fsck", false, "Run a filesystem check on CIPHERDIR")
 	flagSet.BoolVar(&args.one_file_system, "one-file-system", false, "Don't cross filesystem boundaries")
+	flagSet.BoolVar(&args.zerodiriv, "zerodiriv", false, "Create diriv as all-zero files")
 
 	// Mount options with opposites
 	flagSet.BoolVar(&args.dev, "dev", false, "Allow device files")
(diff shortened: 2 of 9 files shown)


Kuroneer:
I'll squash all commits eventually once everything is ok

rfjakob:
I kinda like this, with the exception that you leak the unencrypted sha256 of the plaintext path. That means looking at the gocryptfs.diriv files an attacker can know if you do or do not have the folder `Secret_CIA_Files` (for example) in the top directory.

Can be fixed, but, I wonder if having all diriv all-zero would be good enough for you usecase? That would also fix the problem with the dirivs not being recalculated on mv.

Kuroneer:
> I kinda like this, with the exception that you leak the unencrypted sha256 of the plaintext path. That means looking at the gocryptfs.diriv files an attacker can know if you do or do not have the folder `Secret_CIA_Files` (for example) in the top directory.

That could be solved by salting the sha, either by reverting the change to the root diriv initialization and using the root diriv as salt everywhere or by allowing the cli flag to provide a salt. In this case I'd go with using a random root diriv as salt. I don't believe that having a deterministic diriv for the root is as critical (and what I'm trying to solve here -the issue mentioned in #108- won't be as relevant for the root dir because you usually clone that directory the first time and that's it)

> Can be fixed, but, I wonder if having all diriv all-zero would be good enough for you usecase? That would also fix the problem with the dirivs not being recalculated on mv.

That being said, yes, any kind of deterministic diriv would be good enough. It's a simpler solution, cleaner and wider, but wouldn't it offer very little over `-plaintextnames`? If I understand the behaviour correctly, you'd have the `Secret_CIA_Files` issue everywhere

rfjakob:
Yes salting the sha would fix the leak, multiple ways to do it, should be no problem.

But about the all-zero dirv:

Secret_CIA_Files would still be encrypted, so I don't see how we would leak if you do or do not have it.

rfjakob:
Some background info: i have been thinking about how to have something fully deterministc for a while, best i came up with is
* Aes-siv encryption, all-zero block ivs
* Filename encryption on but with all-zero dirivs

Kuroneer:
> Secret_CIA_Files would still be encrypted, so I don't see how we would leak if you do or do not have it.

Yes, I realized that late, my bad

Kuroneer:
I'll change this PR to have `all-zero diriv` then, not an issue (but it'll be around next week, I'll be AFK for the weekend). And I'll do something to be able to run the tests locally so I don't need to rely on the github actions

rfjakob:
Oh, it's incompatible with earlier gocryptfs versions because of the allzero sanity check.

Change is good, I will merge this, but will extend it to not write the diriv to disk at all. Or do you want to give it a try?

rfjakob:
I'm already working on it.

Kuroneer:
I went with the easiest route, true. I also thought that it'd be better to have compatibility when mounting with or without the flag. I guess previous versions went above my head when I was required to remove the zero iv check.

It opens the following questions:
Should the flag only prevent diriv creation or reading the dirivs altogether? (my guess is that it should only prevent creation to allow the same system to work both with and without the flag)
In that case, missing dirivs should/would be treated as all-zero dirivs, in order to support the previous statement (and the flag would be irrelevant when reading a directory)

rfjakob:
My plan now is to require the flag on `-init`, and store it in gocryptfs.conf, because a user may forget specifying the flag sometimes, and then you'll end up with an unholy mix. Then, not read or write any gocryptfs.diriv files.

rfjakob:
Merged to master, thanks!

dumblob:
> My plan now is to require the flag on `-init`, and store it in gocryptfs.conf, because a user may forget specifying the flag sometimes, and then you'll end up with an unholy mix. Then, not read or write any gocryptfs.diriv files.

It seems the flag `-deterministic-names` is not meant to be the default. I'm new to this, but I'd like to ask why is it not the default?

rfjakob:
Please see https://github.com/rfjakob/gocryptfs/blob/master/Documentation/MANPAGE.md#-deterministic-names

PR #648, opened 2022-03-31
Fix reverse gocryptfs.conf access on macOS

valeriangalliat:
Hey! There's a bug when using gocryptfs on macOS in reverse mode, where we can't access `gocryptfs.conf` in the virtual filesystem (supposed to map to `.gocryptfs.reverse.conf` in the real filesystem). It returns "Permission denied" instead.

FUSE debug logs looks like this when attempting to `cat gocryptfs.conf` in the virtual filesystem:

```
14:44:14.095207 rx 3: GETATTR n2
14:44:14.095229 tx 3:     OK, {tA=1s {M0100000 SZ=0 L=0 0:0 0 0:8954996 A 0.000000 M 0.000000 C 0.000000}}
14:44:14.099943 rx 4: ACCESS n2 {u=501 g=20 r}
14:44:14.099990 tx 4:     13=permission denied
```

I managed to fix it by implementing `Getattr` on `VirtualConfNode`.

Cheers!

### DIFF
--- a/internal/fusefrontend_reverse/virtualconf.go
+++ b/internal/fusefrontend_reverse/virtualconf.go
@@ -10,6 +10,7 @@ import (
 )
 
 var _ = (fs.NodeOpener)((*VirtualConfNode)(nil))
+var _ = (fs.NodeGetattrer)((*VirtualConfNode)(nil))
 
 type VirtualConfNode struct {
 	fs.Inode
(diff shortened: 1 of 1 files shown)


rfjakob:
Wow, how does this even work on Linux

rfjakob:
Huh, it works on Linux because the result of LOOKUP is cached, and GETATTR is never called.

Patch is good, PR description is good, but please update the commit message. That one's a little terse.

valeriangalliat:
Thanks for the review, I just updated the commit message :)

> Huh, it works on Linux because the result of LOOKUP is cached, and GETATTR is never called.

Wow, good finding! Yeah with macFUSE I believe it's the opposite and `GETATTR` gets cached. Typically `ls -la gocryptfs.conf` would give the correct attributes until I try to `cat` the file, then I would get permission denied and the attributes would become all blank from that point on.

rfjakob:
Merged, thanks!

PR #654, opened 2022-04-24
Fix invalid -longnamemax for reverse mode

NekoGirlSAIKOU:
In the reverse mode `-longnamemax` seems not working. Only encrypted file names longer than 255 are hashed.


NekoGirlSAIKOU:
My Calculation for `shortNameMax` seems wrong so I have to replace `log.Panic` to `continue` after

PR #655, opened 2022-04-24
Fix invalid -longnamemax for reverse mode

NekoGirlSAIKOU:
In the reverse mode `-longnamemax` seems not working. Only encrypted file names longer than 255 are hashed.


rfjakob:
Change looks good, but needs a test.

rfjakob:
Test added and merged as https://github.com/rfjakob/gocryptfs/commit/1bff80b46cf3cdd2d6934ebf905fca96dde7af97 , thanks!

PR #705, opened 2023-01-01
resolves #702 add fido2 up/uv/pin support

### LINKED ISSUE #702: RFE: add fido2-with-client-pin
in https://github.com/rfjakob/gocryptfs/issues/571 user verification was removed. as a result currently only user presence is configured. I did not find an option to enable client pin.

I'd like to suggest to add options similar to systemd-cryptenroll
- fido2-device
- fido2-with-client-pin=BOOL, defaul=yes
- fido2-with-user-presence=BOOL, default=yes
- fido2-with-user-verification=BOOL, default=no

the selected options should be added to gocryptfs.conf

Possible solution
depending on the selected option, add `-t [up|uv|pin] `to line https://github.com/rfjakob/gocryptfs/blob/69d88505fd7f4cb0d9e4f1918de296342fe05858/internal/fido2/fido2.go#L44

schaarsc:


rfjakob:
Hi, a few comments:

1) Commit message is missing (text in https://github.com/rfjakob/gocryptfs/issues/702 is good, put it into the commit message)
2) You cannot change existing command-line flags as this breaks existing users
3) The tests fail

schaarsc:
I understand the wish to avoid breaking changes. In this case however, I'd  advocate for it, because a) it provides a cleaner cli and b) I assume this feature is used by tech savvy people that can handle the switch from "--fido" to "--fido-device" if this is documented in the Release Notes.
to make sure already encrypted files are not affected, I left fido2(line 50) untouched in case no parameters are provided. 

As for the code comments... I'm not a go developer, consider this to be a proof of concept, in my opinion a skilled developer should take it from here and make the appropriate changes / quality checks required for this project.

rfjakob:
Functionality will  be merged via the simpler https://github.com/rfjakob/gocryptfs/pull/807

Issue #748, opened 2023-05-20
[Question] Is there a way to use FIDO2 AND have a passwd as fallback?


aanno:

SamBayerZXZ:
I am also interested in this. Hardware key developers insist on registering at least 2 keys in case of loss or failure of one. Yes, it is possible to access using the master key but but there is no way to change the hardware key to another one or change the key to a passphrase. Currently, if you lose your hardware key, you must create a new vault with a new hardware key and migrate data from the old vault to the new one. This is not a problem if the amount of data is small, otherwise it is not convenient. When trying to change the password of a fido2-encrypted storage, we get an error

`gocryptfs -passwd -fido2 /dev/XXX /test/.crypt/`

```
FIDO2 Secret: interact with your device ...
Decrypting master key
Password change is not supported on FIDO2-enabled filesystems.
```

PR #760, opened 2023-06-25
Request for comments: Allow more than one encrypted master key

aanno:
Hello, 

I've hacked a fix to #748. This is _not_ ready for merge. But before putting more work into this, I need some feedback.

So long, `gocryptfs` only allows _one_ instance of the encrypted master key. (Decryption could be done with a passphrase or by using a (modern) FIDO2 device.)

I hacked the `gocryptfs.conf` configuration format from v2 to v3. The new format supports (a) several instances of the encrypted master key and (b) several instances of the FIDO2 device details. v2 could be easily converted to v3.

Omissions:
1. No documentation.
2. Around 9% of the tests still fail (mostly tests that rely on the v2 format).
3. No new tests for the new features.

Usage

```bash
# init a new crypted fs
gocryptfs -init -user user cipher
# add another user/password instance
gocryptfs -user user -add-user testuser cipher
# use it
gocryptfs -user testuser cipher mnt

# add FIDO2 device
gocryptfs -user testuser -add-fido2 /dev/hidraw3 -add-fido2-name solo cipher
# use FIDO2 device
gocryptfs -fido2 /dev/hidraw3 -fido2-name solo cipher mnt
# delete FIDO2 device
gocryptfs -fido2 /dev/hidraw3 -fido2-name solo --delete-fido2-name solo cipher
````

Is that anything `gocryptfs` could consider to merge when ready?

Kind regards,

aanno


rfjakob:
Hi, thanks for the PR! I understand the need, but I am afraid of the complexity and the format change.

Maybe this could be done by having multiple copies of gocryptfs.conf ? Like, gocryptfs.user1.conf etc. This would work already today using the `-config=` flag, but could be made more userfriendly by adding `-user=` as you suggested, though I would call it `keyslot" to match what LUKS does.

schlomie:
This does work already.  The only UX improvement I would recommend is prompting the user for the masterkey (to recreate the config file) instead of accepting it on stdin - as to not expose the masterkey to buffer and history.

PR #807, opened 2023-11-21
Add user FIDO assert option

### LINKED ISSUE #702: RFE: add fido2-with-client-pin

invis-z:
Add an option to specify user verification options for `fido2-assert -t`

Options will be saved to config file

Provide same functionality to #705 with simpler implementation

Resolve #702

### DIFF
--- a/Documentation/MANPAGE.md
+++ b/Documentation/MANPAGE.md
@@ -479,10 +479,16 @@ for details.
 
 #### -fido2 DEVICE_PATH
 Use a FIDO2 token to initialize and unlock the filesystem.
-Use "fido2-token -L" to obtain the FIDO2 token device path.
-For linux, "fido2-tools" package is needed.
+Use `fido2-token -L` to obtain the FIDO2 token device path.
+For linux, **fido2-tools** package is needed.
 
 Applies to: all actions that ask for a password.
+
+#### -fido2-assert-option OPTION
+Options passed to `fido2-assert` with `-t` option.
+This option may be specified multiple times, each time it will add two 
+arguements `-t` `OPTION` to `fido2-assert`.
+See `man fido2-assert` to check supported options.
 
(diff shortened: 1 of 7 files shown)


Documentation/MANPAGE.md
```diff
@@ -479,11 +479,17 @@ for details.
...
+For linux, **fido2-tools** package is needed.
 
 Applies to: all actions that ask for a password.
 
+#### -fido2-assert-option OPTION
+Options passed to `fido2-assert` with `-t` option.
+This option may be specified multiple times, each time it will add two 
+arguements `-t` `OPTION` to `fido2-assert`.
+See `man fido2-assert` to check supported options.
+
```
rfjakob, line 492: An example would be nice here: create a filesystem and mount. With fido2 and fido2-assert-option.

rfjakob (approved):

rfjakob:
Code change looks good, I just need to test this locally. Which involves finding my fido2 token again.

rfjakob:
I still did not get to testing myself, and I don't want to block this any longer, so I'm gonna merge now and trust that it works.

rfjakob:
Thanks!

PR #831, opened 2024-03-09
init_dir: use masterkey arg

### LINKED ISSUE #824: directory init and masterkey arg
Currently if the `-masterkey` argument is passed when initializing a new directory, this is not used and a random one is generated instead. It would be nice not to skip this argument and let the user define the `masterkey` when running the init command.

pmazzini:
Fixes #824

### DIFF
--- a/cli_args.go
+++ b/cli_args.go
@@ -280,7 +280,7 @@ func parseCliOpts(osArgs []string) (args argContainer) {
 		tlog.Fatal.Printf("The options -passfile and -masterkey cannot be used at the same time")
 		os.Exit(exitcodes.Usage)
 	}
-	if len(args.extpass) > 0 && args.masterkey != "" {
+	if len(args.extpass) > 0 && args.masterkey != "" && !args.init {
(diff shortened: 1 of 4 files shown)


rfjakob:
Code change is good, but needs a test. Can you add TestInitMasterkey similar to https://github.com/rfjakob/gocryptfs/blob/master/tests/cli/cli_test.go#L77 ?

rfjakob (changes requested):
Code change is good, test is needed

cli_args.go
```diff
@@ -280,10 +280,6 @@ func parseCliOpts(osArgs []string) (args argContainer) {
 		tlog.Fatal.Printf("The options -passfile and -masterkey cannot be used at the same time")
 		os.Exit(exitcodes.Usage)
 	}
-	if len(args.extpass) > 0 && args.masterkey != "" {
```
pmazzini, line 283: From https://github.com/rfjakob/gocryptfs/commit/295d432175292dbaef572093d784aab55f5c0b8f I don't see why can't these 2 be used together.

rfjakob:
Merged, thanks!

Issue #557, opened 2021-03-25
gocryptfs support to specify the volume name?


corelanc0d3r:
Hi,

I was wondering if there is a way to specify the desired volume name when mounting a gocryptfs folder?
Assuming gocryptfs was accepting similar syntax as encfs, I tried to pass the "volname" flag via -o ... but then I get this:

```
gocryptfs /path/to/cipherfolder /path/to/mountfolder -o volname="blah"
flag provided but not defined: -volname
Invalid command line: gocryptfs -volname=blah /path/to/cipherfolder /path/to/mountfolder. Try 'gocryptfs -help'.
```

What is the correct syntax to use?

thanks !

rfjakob:
The flag is called `fsname`, and it needs to be passed to the kernel, so you have to use `-ko` (kernel option). This works:

    gocryptfs -ko fsname=blah cipher mnt

corelanc0d3r:
Unfortunately it didn't seem to work for me:

```
corelanc0d3r@/Users/corelanc0d3r/~ % gocryptfs -ko fsname="blah" CTest /Users/corelanc0d3r/mnt/test 
Password: 
Decrypting master key
Filesystem mounted and ready.
corelanc0d3r@/Users/corelanc0d3r/ ~ % 
```
Volume was mounted and shows up as "test" (name of the mount folder), not as "blah"

rfjakob:
Wow, looks like this is called "fsname" on Linux and "volname" on MacOS (as you said!). I only tested on Linux.

Can you try again with

gocryptfs -ko volname="blah"

?

corelanc0d3r:
Unfortunately that didn't work either

I can confirm however that it works when mounting volumes using encfs, but perhaps encfs is using macFuse, and gocryptfs is using osxfuse?
please don't hesitate to let me know what I can do to help troubleshooting the issue

thanks !

mike-kfed:
[...]

PR #855, opened 2024-06-15
cli: deduplicate kernel options

### LINKED ISSUE #854: Avoid duplicate kernel options
This is a follow-up to #557. I would like to know if the proposed changes are acceptable.

The `-ko volname=XXX` does not work, because gocryptfs sets `volname` to directory base name of the mountpoint first, then expect to override it with duplicate option-value pair. Relevant codes:

https://github.com/rfjakob/gocryptfs/blob/f06f27e7bc098e334024c365004f9303e79997d9/mount.go#L437-L442

https://github.com/rfjakob/gocryptfs/blob/f06f27e7bc098e334024c365004f9303e79997d9/mount.go#L467-L473

Basically, it would produce `-o volname=BaseNameOfMountPoint,volname=CustomVolName` option for macFUSE.
This appears logical at first glance, but it doesn't work as expected. macFUSE only reads the initial `volname` value set by gocryptfs and does not check for updates to this option later on.

I swapped these two code blocks and test on macOS Sonoma and it works as expected, the `volname` option can now be overridden correctly.

I think it's worth rewriting this part of gocryptfs so it only sets each option exactly once - use only user-provided option if possible, only inject stock options if they are not provided.

JokerQyou:
Merge stock kernel options with user-provided ones before passing to go-fuse.

Before: `-ko volname=custom` would result in `-o volname=mountpoint,volname=custom` to macFUSE.

After: `-ko volname=custom` would produce `-o volname=custom` with no duplicates.

Fixes #854 and #557

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -402,12 +402,13 @@ func initGoFuse(rootNode fs.InodeEmbedder, args *argContainer) *fuse.Server {
 	}
 
 	mOpts := &fuseOpts.MountOptions
+	opts := make(map[string]string)
 	if args.allow_other {
 		tlog.Info.Printf(tlog.ColorYellow + "The option \"-allow_other\" is set. Make sure the file " +
 			"permissions protect your data from unwanted access." + tlog.ColorReset)
 		mOpts.AllowOther = true
 		// Make the kernel check the file permissions for us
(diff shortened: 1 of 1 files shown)


rfjakob:
Thanks!

PR #878, opened 2024-11-18
Xattrs support in reverse mode

Ratio2:
Fixes https://github.com/rfjakob/gocryptfs/issues/827


rfjakob:
Nice! Can you enable the test in https://github.com/rfjakob/gocryptfs/blob/master/tests/reverse/xattr_test.go ?

internal/fusefrontend/node_xattr.go
```diff
@@ -65,15 +62,11 @@ func (n *Node) Getxattr(ctx context.Context, attr string, dest []byte) (uint32,
...
 		if err != nil {
 			tlog.Warn.Printf("GetXAttr: %v", err)
-			return minus1, syscall.EIO
+			return 0, syscall.EIO
 		}
 	}
-	// Caller passes size zero to find out how large their buffer should be
-	if len(dest) == 0 {
-		return uint32(len(data)), 0
-	}
```
rfjakob, line 74: Why remove this?

internal/fusefrontend/node_xattr.go
```diff
@@ -50,13 +47,13 @@ func (n *Node) Getxattr(ctx context.Context, attr string, dest []byte) (uint32,
 		var errno syscall.Errno
 		data, errno = n.getXAttr(attr)
 		if errno != 0 {
-			return minus1, errno
```
rfjakob, line 53: Why not -1 like https://man7.org/linux/man-pages/man2/getxattr.2.html#RETURN_VALUE says?

Ratio2: The behavior of the function is described in:
https://github.com/hanwen/go-fuse/blob/756578b6a0b03511fde147b2f23c5b97217ede27/fs/api.go#L312
Returning -1 does not make sense, since it is ignored on errors:
https://github.com/hanwen/go-fuse/blob/756578b6a0b03511fde147b2f23c5b97217ede27/fuse/opcode.go#L294

internal/fusefrontend/node_xattr.go
Ratio2: The behavior of the function is described in:
https://github.com/hanwen/go-fuse/blob/756578b6a0b03511fde147b2f23c5b97217ede27/fs/api.go#L312
It doesn't make sense, since the following condition will give the same result:
https://github.com/hanwen/go-fuse/blob/756578b6a0b03511fde147b2f23c5b97217ede27/fuse/opcode.go#L285

rfjakob:
Merged as https://github.com/rfjakob/gocryptfs/commit/ed1c5e4a9f5ce1921f3ec03b32e591ce828ec5b9 , thanks!

PR #935, opened 2025-07-10
remove redundant contentEnc reference from file struct

lquenti:
`File` only gets created via `NewFile()` with the reference from the root node, but each file has a ref on the root node itself as well, thus its redundant.

[tests look good I guess](https://github.com/user-attachments/files/21165262/testoutput.txt)

[See here for related discussion](https://github.com/rfjakob/gocryptfs/discussions/933)


rfjakob:
Thanks!

PR #938, opened 2025-07-15
The master key is printed at initialization, not on mount

lquenti:
Maybe I missunderstand the docs, but the master key is printed out only when initially creating it, especially with the note that

> This message is only printed once.


rfjakob:
Right! That has changed a long time ago and the text was not updated. Thanks!

PR #939, opened 2025-07-15
feat: Add  and  commands

dzianisv:
This commit introduces two new commands:

- `gocryptfs -takeout CIPHERDIR PATH DESTDIR`: Decrypts a specific file or directory from the encrypted filesystem and moves it to a destination, deleting the original encrypted file.
- `gocryptfs -list CIPHERDIR`: Lists the decrypted file and directory names in a tree-like structure.

These commands enhance the utility of `gocryptfs` by providing more granular control over file decryption and better visibility into the encrypted filesystem's contents without requiring FUSE.

fix: `list` command outputs full decrypted file paths

The previous implementation of the `list` command outputted a tree-like structure and failed to decrypt special files. This commit modifies the `list` command to:

- Output full, relative decrypted file paths, similar to `git ls-files`.
- Correctly handle and skip special files like `gocryptfs.diriv` and `gocryptfs.conf`.
- Decrypt each component of the path individually to ensure correct decryption of nested directories.

fix: `takeout` command handles plaintext paths correctly

The `takeout` command previously failed when provided with a plaintext path, as it attempted to directly access the encrypted equivalent. This commit refactors the `takeout` command to:

- Walk the entire encrypted `CIPHERDIR`.
- Decrypt each file's relative path on the fly.
- Compare the decrypted path with the user-provided plaintext `PATH`.
- Only process files and directories that match or are children of the specified `PATH`.
- Add a helper function `decryptRelativePath` to encapsulate the path decryption logic.
- Skip files starting with `._` to avoid decryption errors with macOS metadata files.

fix: `takeout` command handles longnames and special files

The `takeout` command previously failed when encountering longnames (hashed filenames) and special files (like macOS metadata files starting with `._`). This commit addresses these issues by:

- Adding `IsLongName` and `ReadLongName` functions to `internal/nametransform/names.go` to correctly identify and read the original plaintext names of longname files.
- Modifying `take_out.go` to use these new functions, ensuring that longnames are properly decrypted.
- Expanding the special file skipping logic in `take_out.go` to include files starting with `._`, preventing decryption errors with macOS metadata.
- Gracefully handling decryption errors for non-gocryptfs files by logging them at a debug level instead of warnings.

fix: `takeout` and `list` commands handle non-gocryptfs filenames

The `takeout` and `list` commands previously failed when encountering filenames that were not valid `gocryptfs` encrypted names (e.g., non-encrypted files, macOS metadata, or files copied without `gocryptfs`'s knowledge). This commit addresses these issues by:

- Adding `IsValidBase64` function to `internal/nametransform/names.go` to check if a string is a valid base64 encoding.
- Modifying `decryptRelativePath` in `take_out.go` and `list.go` to use `IsValidBase64`. If a filename component is not a valid base64 string, it is now treated as a literal plaintext name and passed through without decryption. This prevents "bad message" and "padding too long" errors.
- Ensuring that only properly encrypted filenames are subjected to decryption, making the tools more resilient to unexpected file types.$

### DIFF
--- a/cli_args.go
+++ b/cli_args.go
@@ -31,7 +31,7 @@ type argContainer struct {
 	longnames, allow_other, reverse, aessiv, nonempty, raw64,
 	noprealloc, speed, hkdf, serialize_reads, hh, info,
 	sharedstorage, fsck, one_file_system, deterministic_names,
-	xchacha bool
+	xchacha, takeout, list bool
 	// Mount options with opposites
 	dev, nodev, suid, nosuid, exec, noexec, rw, ro, kernel_cache, acl bool
 	masterkey, mountpoint, cipherdir, cpuprofile,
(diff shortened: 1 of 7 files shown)


rfjakob:
Hi, uh, first, is this AI generated? Please declare this in the PR already.

These are big features and will need to go big in tests. I don't see a single one right now.

PR #949, opened 2025-07-20
Fix macos filename encoding

rafjaf:
I rewrote PR https://github.com/rfjakob/gocryptfs/pull/936 (where more information is given) to adapt it to version 2.6.0 (the dirstream fix had become obsolete since switching to the new go-fuse directory API).

----

From #936:

This PR aims at solving the issue described in https://github.com/rfjakob/gocryptfs/issues/850

Since I am no professional programmer, I used Claude Sonnet 4 to try to implement the solution used in Cryptomator to solve a similar problem. In course of this PR, another problem was discovered regarding the implementation of dirstream on macOS, and I tried to solve it also.

For the sake of transparency, here is a detailed summary of the changes implemented in this PR - I hope it's not too verbose but I thought it might help better understand the logic behind the PR and allow you correct it if need be:

# macOS Compatibility Improvements for gocryptfs

This document describes the major changes made to improve gocryptfs compatibility on macOS, addressing critical issues with Unicode filename handling and directory listing functionality.

### Problem Statement

On macOS, there is a fundamental mismatch between how different applications handle Unicode normalization for filenames containing accented characters (like "café"):

- **Command-line tools** (ls, touch, cat, etc.) typically use **NFC** (Canonical Composed) form: `café` = `c` + `a` + `f` + `é` (4 characters, where `é` is U+00E9)
- **GUI applications** (Finder, TextEdit, etc.) use **NFD** (Canonical Decomposed) form: `café` = `c` + `a` + `f` + `e` + `́` (5 characters, where `é` = `e` + combining acute accent U+0301)

This caused serious usability problems in gocryptfs on macOS:

1. **Duplicate files**: A file created via CLI as "café.txt" (NFC) would appear different from "café.txt" created via Finder (NFD), even though they look identical to users
2. **Invisible files**: Files created by CLI tools would not appear in Finder, and vice versa
3. **Access failures**: TextEdit couldn't open files created by `touch`, and `cat` couldn't read files created by Finder
4. **Broken workflows**: Users couldn't seamlessly work with the same files using both GUI and CLI tools

### Solution: Cryptomator-Inspired Approach

The implementation adopts the approach used by [Cryptomator](https://github.com/cryptomator/cryptomator), specifically described [here](https://github.com/cryptomator/cryptomator/issues/264):

#### Core Principles
1. **Canonical Storage**: Always store filenames internally in NFC (composed) form for consistency
2. **Transparent Lookup**: When looking up a file, try both NFC and NFD forms to find existing files
3. **Migration**: If an NFD file exists but no NFC equivalent, migrate the NFD file to NFC
4. **Platform-Appropriate Display**: Return directory listings in the form expected by the OS (NFD on macOS for GUI compatibility)

#### Algorithm Flow
```
File Creation (Create, Mkdir, Symlink, etc.):
Input filename → Normalize to NFC → Store with NFC name

File Lookup (Lookup, Open, Stat, etc.):
Input filename → Try NFC lookup
  ↓ If found: return file
  ↓ If not found: try NFD lookup
    ↓ If NFD found: migrate NFD→NFC, return file
    ↓ If neither found: return ENOENT

Directory Listing (Readdir):
Stored NFC names → Convert to NFD for display on macOS
```

#### Normalization Functions
Added core normalization functions in `internal/fusefrontend/node_dir_ops.go`:

```go
// normalizeFilename converts filenames to NFC for consistent internal storage
func normalizeFilename(name string) string {
    if runtime.GOOS == "darwin" && utf8.ValidString(name) {
        return norm.NFC.String(name)
    }
    return name
}

// normalizeFilenameForDisplay converts NFC to NFD for macOS GUI compatibility
func normalizeFilenameForDisplay(name string) string {
    if runtime.GOOS == "darwin" && utf8.ValidString(name) {
        return norm.NFD.String(name)
    }
    return name
}
```

#### File Creation Operations
Updated all file creation operations to normalize input to NFC:

- **Files**: `internal/fusefrontend/node_open_create.go`
- **Directories**: `internal/fusefrontend/node_dir_ops.go`
- **Symlinks, Device nodes, Hard links**: `internal/fusefrontend/node.go`

#### Lookup with Fallback and Migration
The core innovation is in `internal/fusefrontend/node_prepare_syscall.go`. Completely rewrote the `prepareAtSyscall` function to implement Cryptomator's lookup logic:

```go
func (n *Node) prepareAtSyscall(child string) (dirfd int, cName string, errno syscall.Errno) {
    if runtime.GOOS == "darwin" && utf8.ValidString(child) {
        // Step 1: Always try NFC first (canonical storage form)
        normalizedChild := norm.NFC.String(child)
        dirfd, cName, errno = n.prepareAtSyscallDirect(normalizedChild)
        if errno == 0 {
            return dirfd, cName, 0  // Found NFC version
        }
        
        // Step 2: Try alternate form if input was different
        if normalizedChild != child {
            // Input was NFD, try original NFD form
            dirfdNFD, cNameNFD, errnoNFD := n.prepareAtSyscallDirect(child)
            if errnoNFD == 0 {
                // Found NFD file - migrate it to NFC
                if errno := n.migrateFilename(child, normalizedChild); errno == 0 {
                    return n.prepareAtSyscallDirect(normalizedChild)  // Use migrated NFC
                } else {
                    return dirfdNFD, cNameNFD, 0  // Migration failed, use NFD
                }
            }
        }
        
        // Step 3: If input was NFC, also try NFD as fallback
        if normalizedChild == child {
            nfdChild := norm.NFD.String(child)
            if nfdChild != child {
                dirfdNFD, cNameNFD, errnoNFD := n.prepareAtSyscallDirect(nfdChild)
                if errnoNFD == 0 {
                    // Found NFD file - migrate it to NFC
                    if errno := n.migrateFilename(nfdChild, normalizedChild); errno == 0 {
                        return n.prepareAtSyscallDirect(normalizedChild)
                    } else {
                        return dirfdNFD, cNameNFD, 0
                    }
                }
            }
        }
    }
    
    return n.prepareAtSyscallDirect(child)  // Non-macOS or fallback
}
```

#### Migration Logic
The `migrateFilename` function handles moving NFD files to NFC:

```go
func (n *Node) migrateFilename(oldName, newName string) syscall.Errno {
    if oldName == newName {
        return 0  // Nothing to do
    }
    
    // For plaintext names: simple rename
    if rn.args.PlaintextNames {
        return syscallcompat.Renameat(dirfd, oldName, dirfd, newName)
    }
    
    // For encrypted names: encrypt both names and rename
    return syscallcompat.Renameat(dirfd, oldCName, dirfd, newCName)
}
```

#### Directory Listing
Updated `Readdir` in `internal/fusefrontend/node_dir_ops.go` to return NFD for macOS GUI compatibility:

```go
func (n *Node) Readdir(ctx context.Context) (fs.DirStream, syscall.Errno) {
    // ... read and decrypt directory entries ...
    
    // Convert from storage format (NFC) to OS-expected format
    name = normalizeFilenameForDisplay(name)  // NFC → NFD on macOS
    
    // ... return directory entries ...
}
```

### Files Modified
- `internal/fusefrontend/node_prepare_syscall.go`: Core lookup and migration logic
- `internal/fusefrontend/node_dir_ops.go`: Normalization functions and directory operations
- `internal/fusefrontend/node_open_create.go`: File creation normalization
- `internal/fusefrontend/node.go`: Symlink/device/hardlink normalization

### Problem in Reverse Mode

Reverse mode presents an encrypted view of plaintext files. The Unicode normalization problem affects how encrypted filenames are decrypted back to plaintext names that must match existing files on disk.

After decrypting an encrypted filename, the resulting plaintext name might not match the actual file on disk due to Unicode normalization differences:
- Encrypted name decrypts to `café` (NFC form)
- But actual file on disk is `café` (NFD form) - created by Finder
- Result: File not found error, even though the file exists

### Solution Approach

Unlike forward mode, **reverse mode should not modify the plaintext filesystem**. Instead, we implement **fallback lookup logic**: if the decrypted name doesn't exist on disk, try the alternate Unicode normalization form.

#### Key Principles
- **No migration**: Don't rename plaintext files (they might be used by other applications)
- **Fallback only**: Try both NFC and NFD forms when looking up files
- **Read-only approach**: Reverse mode remains a presentation layer without modifying source files

#### Enhanced `rDecryptName` Function
In `internal/fusefrontend_reverse/rpath.go`, added Unicode normalization fallback logic:

```go
func (rfs *RootNode) rDecryptName(cName string, dirIV []byte, pDir string) (pName string, err error) {
    // ... existing decryption logic ...
    
    // NEW: On macOS, handle Unicode normalization fallback
    if runtime.GOOS == "darwin" && utf8.ValidString(pName) {
        // Check if the decrypted name actually exists on disk
        pPath := filepath.Join(rfs.args.Cipherdir, pDir, pName)
        var st syscall.Stat_t
        if statErr := syscall.Stat(pPath, &st); statErr != nil {
            // Try the alternate Unicode form
            var alternateName string
            if norm.NFC.String(pName) == pName {
                // pName is NFC, try NFD
                alternateName = norm.NFD.String(pName)
            } else {
                // pName is NFD (or mixed), try NFC
                alternateName = norm.NFC.String(pName)
            }
            
            if alternateName != pName {
                alternatePath := filepath.Join(rfs.args.Cipherdir, pDir, alternateName)
                var altSt syscall.Stat_t
                if altStatErr := syscall.Stat(alternatePath, &altSt); altStatErr == nil {
                    // The alternate form exists, use it
                    return alternateName, nil
                }
            }
        }
    }
    
    return pName, nil
}
```

### Key Differences from Forward Mode

| Aspect | Forward Mode | Reverse Mode |
|--------|-------------|--------------|
| **Storage Control** | ✅ Can rename encrypted files | ❌ Cannot modify plaintext files |
| **Migration** | ✅ Migrate NFD→NFC | ❌ No migration (read-only) |
| **Approach** | Enforce NFC storage + migrate | Fallback lookup only |
| **Complexity** | High (full Cryptomator logic) | Low (simple fallback) |

### Files Modified
- `internal/fusefrontend_reverse/rpath.go`: Added Unicode imports and enhanced `rDecryptName()` with fallback lookup logic
- `internal/fusefrontend_reverse/node_dir_ops.go`: Added Unicode imports and helper functions

### Problem Discovery

During the implementation of Unicode normalization, we discovered a critical bug in directory listing functionality. While file operations worked correctly, directory listings (via `ls`, Finder, etc.) would show empty directories even when files were present. This might be the same problem as the "ghost" mountpoint mentioned in https://github.com/rfjakob/gocryptfs/issues/898

#### Symptoms
- Files could be created, read, written, and deleted successfully
- File operations via direct path access worked normally
- Directory listings (`ls`, `find`, Finder) showed empty directories
- The underlying encrypted directory contained the expected files
- Raw Go directory reading worked correctly

### Root Cause Analysis

The issue was traced to a presumed incompatibility between the go-fuse library's `NewLoopbackDirStreamFd` function and macOS/APFS filesystem behavior:

1. **Raw directory reading worked**: Using Go's standard `os.File.Readdirnames()` could successfully read directory entries
2. **go-fuse loopback stream failed**: `fs.NewLoopbackDirStreamFd()` consistently returned `nil` entries immediately, indicating "end of directory"
3. **File descriptor ownership issues**: The loopback stream appeared to take ownership of file descriptors in a way that conflicted with macOS directory access patterns

#### Why This Affects macOS Specifically
1. **APFS vs ext4/xfs differences**: macOS uses APFS which has different directory iteration semantics compared to Linux filesystems
2. **Darwin syscall differences**: macOS implements POSIX directory operations with subtle differences that affect low-level libraries
3. **go-fuse library testing**: The go-fuse library is primarily developed and tested on Linux, with less comprehensive macOS testing

### Solution: Platform-Specific Directory Stream

Implemented a custom directory stream that is only used on macOS (`runtime.GOOS == "darwin"`), while maintaining the standard go-fuse loopback implementation on other platforms.

#### Implementation
```go
// Use custom directory stream on macOS due to issues with go-fuse loopback implementation
if runtime.GOOS == "darwin" {
    // On macOS, use our custom directory stream implementation
    // The go-fuse NewLoopbackDirStreamFd has compatibility issues with macOS/APFS
    ds = &customDirStream{fd: fdDup}
    errno = 0
} else {
    // On other platforms, use the standard loopback directory stream
    ds, errno = fs.NewLoopbackDirStreamFd(fdDup)
    if errno != 0 {
        goto err_out
    }
}
```

#### Custom Directory Stream Features
The `customDirStream` struct implements all required interfaces:
- `fs.DirStream` - Core directory stream interface
- `fs.FileReaddirenter` - For reading directory entries
- `fs.FileSeekdirer` - For seeking within directory streams
- `fs.FileReleasedirer` - For cleanup
- `fs.FileFsyncdirer` - For sync operations

Key implementation details:
1. **Lazy Loading**: Directory entries are loaded on first access using `os.File.Readdirnames(-1)`
2. **Proper File Descriptor Management**: Uses the provided file descriptor without taking improper ownership
3. **Seeking Support**: Implements proper seek behavior for directory offsets
4. **Interface Compliance**: Implements all required methods to be a drop-in replacement

### Files Modified
- `internal/fusefrontend/file_dir_ops.go`: Added `runtime` import, modified `OpendirHandle()` to use platform-specific directory streams, implemented `customDirStream` type

## 4. Platform Compatibility

All implementations are macOS-specific (guarded by `runtime.GOOS == "darwin"`) because:
- Linux typically uses UTF-8 without normalization issues
- Windows uses UTF-16 with different Unicode handling
- Only macOS has the CLI/GUI normalization split that causes this problem
- The directory stream issue only affects macOS/APFS

### DIFF
--- a/tests/macos_filename_encoding/nfc_nfd_test.go
+++ b/tests/macos_filename_encoding/nfc_nfd_test.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"golang.org/x/text/unicode/norm"
@@ -13,6 +14,20 @@ import (
 
 var nfcName = norm.NFC.String("e\u0301") // é
 var nfdName = norm.NFD.String("e\u0301") // e + combining acute accent
+
+// Additional test cases with various Unicode characters
+var unicodeTestCases = []struct {
+	name    string
+	nfc     string
+	nfd     string
+	comment string
+}{
+	{"acute_e", norm.NFC.String("e\u0301"), norm.NFD.String("e\u0301"), "é (e + combining acute)"},
+	{"circumflex_a", norm.NFC.String("a\u0302"), norm.NFD.String("a\u0302"), "â (a + combining circumflex)"},
+	{"tilde_n", norm.NFC.String("n\u0303"), norm.NFD.String("n\u0303"), "ñ (n + combining tilde)"},
+	{"umlaut_u", norm.NFC.String("u\u0308"), norm.NFD.String("u\u0308"), "ü (u + combining diaeresis)"},
+	{"multiple_combining", norm.NFC.String("o\u0302\u0308"), norm.NFD.String("o\u0302\u0308"), "ô̈ (o + circumflex + diaeresis)"},
+}
 
 func TestCreateNFC_AccessNFD(t *testing.T) {
 	if runtime.GOOS != "darwin" {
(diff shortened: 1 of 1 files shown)


rfjakob:
Traveling right now, can only review in detail in 2 weeks.

Some points already:
* Commit message needs work
* Please drop the migrate logic for now, this is too risky
* Needs tests

rafjaf:
Thanks, I have already changed the commit message and will implement the other changes asap

rafjaf:
@rfjakob As requested, I have:
- rewritten the commit message to make it more explicit
- Dropped the migration logic
- Added testing in /tests/macos_filename_encoding

Also, after further reflection, I have cancelled all changes that I had made to the reverse mode. Indeed, in reverse mode, the plaintext folder is just an ordinary folder, meaning that macos is already doing the normalization required; and as far as I can tell, the existing version of gocrypts has no problem to encode both NFC and NFD filenames to the ciphertext folder. The PR is therefore much simplified.

I hope this works for you and remain at your disposal to discuss this.
Thank you for your time

rafjaf:
Sorry, I had forgotten to revert the changes in /internal/fusefrontend_reverse/node_dirs_ops.go, now all the changes in reverse mode should effectively be cancelled.

rafjaf:
@rfjakob Just to be sure, do you wish me to implement all or part of Copilot's suggestions, or do you prefer to take over from here?

rfjakob:
> @rfjakob Just to be sure, do you wish me to implement all or part of Copilot's suggestions, or do you prefer to take over from here?

I'll take over

rafjaf:
Dear @rfjakob I remain at your disposal if I can be of help to finalize this pull request

rfjakob:
Background info:

* https://github.com/rfjakob/gocryptfs/pull/936
* https://eclecticlight.co/2021/05/08/explainer-unicode-normalization-and-apfs/
* https://github.com/cryptomator/cryptomator/issues/264

rafjaf:
@rfjakob Out of curiosity, I had a look at your two most recent commits. Thank you for continuing to work on this.

I may be missing something, but as I understand the current changes, they establish a clear convention whereby filenames are normalized to NFC internally, while being presented as NFD to user-space (e.g. via readdir).

What I am unsure about is whether this fully replaces the earlier idea discussed in this PR regarding the prepareAtSyscall function, which used to attempt to resolve a path component first as NFC and, if that fails, as NFD as a fallback. My concern is that different Mac applications still appear to pass filenames in different normalization forms (CLI vs GUI), and that accepting only a single normalization form at the syscall boundary might still lead to edge cases. This may also be problematic if a vault already contains "legacy" entries whose effective plaintext name is stored in NFD (created under older behavior). 

I may have misunderstood the intent or the scope of the current commits, so I’d be very happy to be corrected. I just wanted to double-check whether this aspect was intentionally dropped or whether it is/will be handled elsewhere now.

rfjakob:
Yes I dropped the fallback intentionally. I did not understand why we would need it when all (new) files are normalized.

Could you test that branch?

rfjakob:
Old files are a problem, yes. They would become inaccessible. Still thinking about it.

rafjaf:
> Yes I dropped the fallback intentionally. I did not understand why we would need it when all (new) files are normalized.
> 
> Could you test that branch?

@rfjakob So, here is what I did to test the latest commit (just to make sure this is what I was supposed to do):
```
git clone https://github.com/rfjakob/gocryptfs.git
cd gocryptfs
git fetch origin a753e0075e854f57689b2b6efa00a447149df8b9
git checkout a753e0075e854f57689b2b6efa00a447149df8b9
```
and then building the app after confirming that ‎internal/fusefrontend/node_prepare_syscall.go did not contain the fallback mechanism.

Well, the result is unsatisfying:
- Previously created files in the mounted dir which contained accents in their filename have become invisible, both in the CLI and in the Finder
- Newly created files, either in the CLI or in the Finder, which contain accents in their filename, are invisible both in the CLI and in the Finder.

When I say invisible, it is a different behaviour than the base version of gocryptfs : in the base version, the files with accents were visible both in CLI and Finder, but they could only be opened/modified in the CLI, while the Finder listed them but it was impossible to act on them (File not found).

This seems to argue for keeping a fallback mechanism, even a simplified one.

Happy to discuss this!

rfjakob:
Thanks for testing! Interesting. Seem like the "fallback mechanism" was actually the only codepath this worked.

Let me check if I see other bugs...

rfjakob:
Fixed via https://github.com/rfjakob/gocryptfs/commit/5185e92ec80a9998ae00cb799dfdfbb93247dfaf instead

PR #984, opened 2025-12-18
Feature: offer to create mountpoint if it does not exist

llopezmo-rh:
Hi.

I have used Gocryptfs many times and I have missed that the mount directory, often temporary, is created automatically in case it does not exist. That is why I am suggesting the option `mkdir` here, if you consider it suitable.

If have tested it and it seems to work fine:

~~~
$ ./gocryptfs -mkdir ciphdir /tmp/x
Mountpoint "/tmp/x" created.
Password:
~~~

If you would like to change anything or you have any preference, please feel free to let me know. For example, I have used `mkdirAll` , but a more cautious option would be `mkdir` instead.

I am an amateur developer and I am not a Go expert. In case something is wrong, I am available to correct it.

Kind regards,
Lucas.

### DIFF
--- a/cli_args.go
+++ b/cli_args.go
@@ -31,7 +31,7 @@ type argContainer struct {
 	longnames, allow_other, reverse, aessiv, nonempty, raw64,
 	noprealloc, speed, hkdf, serialize_reads, hh, info,
 	sharedstorage, fsck, one_file_system, deterministic_names,
-	xchacha bool
+	xchacha, mkdir bool
 	// Mount options with opposites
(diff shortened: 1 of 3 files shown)


llopezmo-rh:
Hi, @rfjakob.

I am available if you would like to check or correct this. Feel free to share your thoughts.

Thank you.

rfjakob:
I'm reluctant to add this as it can be easily accomplished in shell:

MNT=$(mktemp -d) ; gocryptfs xyz $MNT

PR #988, opened 2026-01-05
Update validate.go

DMyachin:
Small format in File content encryption validation


rfjakob:
Code churn but advantage unclear

PR #992, opened 2026-01-23
macos: normalize unicode file names: store as NFC, read as NFD

rfjakob:
This commit resolves https://github.com/rfjakob/gocryptfs/issues/850 by addressing Unicode normalization mismatches on macOS between NFC (used by CLI tools) and NFD (used by GUI apps). The solution is inspired by Cryptomator's approach ( https://github.com/cryptomator/cryptomator/issues/264 ).

Forward mode on MacOS now enforces NFC for storage but presents NFD as recommended by https://developer.apple.com/library/archive/qa/qa1173/_index.html and https://github.com/macfuse/macfuse/wiki/File-Names-(Unicode-Normalization-Forms) .

See https://github.com/rfjakob/gocryptfs/pull/949 for more info.


rfjakob:
@rafjaf can you try this one?

rafjaf:
Dear @rfjakob, now it works much better : if I create a file with accents in the name, whether in the CLI or in a GUI, it can be opened, both in the CLI or in a GUI. Thank you very much, your solution is much more elegant than the one I had proposed!

The only remaining problem is if the vault contains preexisting files whose name contains accents encoded in NFD format by  gocryptfs 2.6.1 or earlier (because they were created in the GUI). In that case, under PR992, these files become invisible, they are not listed anymore in the CLI (ls) or a GUI (Finder). I am not sure exactly why, byt unless a more sophisticated solution is implemented to solve this legacy problem,  perhaps this should be documented somewhere so that Mac users know what to expect.

rfjakob:
Great!

How does it look with `-plaintextnames`?

rafjaf:
@rfjakob It also works with `-plaintextnames` : if I init a new vault and mount it with this option, and create a file with accented filename in this new vault, with PR992, I can open it, both in the CLI and in the GUI, both in the cipherdir and in the mountdir. If I do the same with the base version of gocryptfs, the filename in the cipherdir remains encrypted and does not correspond to the accented filaname in the mounted dir.

PR #996, opened 2026-02-09
enable falloc on btrfs if CoW is disabled

cebtenzzre:
Currently, gocryptfs disables prealloc on all btrfs mounts, because COW semantics prevent it from working as intended.

However, btrfs makes it straightforward to disable COW on the backing directory with `chattr +C` or  by mounting the device with `nodatacow`. This causes extents to be updated in-place like other filesystems, allowing falloc to work as expected.

This PR prevents gocryptfs from automatically disabling prealloc when this is detected.

As a heuristic, we only check the nocow flag on the backing directory itself, and assume files within inherit this flag, which should be the case as long as the flag was set before any files were created inside.


rfjakob:
WITH A TEST ❤️

Merged as b024e3696d215e6b7a0b8d17ca3eb28af20dd504 , thanks!

PR #1005, opened 2026-04-18
mount: mount ro if cipherdir is on a ro filesystem (don't allow rw)

SimonPilkington:
Currently, if the cipherdir being mounted is on a readonly filesystem, it will be mounted `rw` (will appear with the `rw` mount option in `/proc/self/mountinfo`) but still return `EROFS` on write attempt. This confuses some software (e.g. KDE's Dolphin).

Fix this by forcing a `ro` mount in this case.

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -449,9 +449,18 @@ func initGoFuse(rootNode fs.InodeEmbedder, args *argContainer) *fuse.Server {
 	if runtime.GOOS == "darwin" {
 		opts["volname"] = strings.Replace(path.Base(args.mountpoint), ",", "_", -1)
 	}
+	underlyingFilesystemRo, err := isReadOnlyFilesystem(args.cipherdir)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


rfjakob:
Are you sure you need the /proc/self/mountinfo thing?

rfjakob:
And downgrade the message to Info, please

SimonPilkington:
> Are you sure you need the /proc/self/mountinfo thing?

I'd been led to believe statfs() by itself would miss a readonly bind mount but upon testing this is not the case. Good catch. Dropped that part.

>And downgrade the message to Info, please

Yes, it looks like the CI doesn't like the warning. Done.

SimonPilkington:
For the record reading the kernel source[1] confirms that statfs does indeed check both mount and superblock flags. Should've just done this to begin with.

[1] https://github.com/torvalds/linux/blob/897d54018cc9aa97fd1529ca08a53b429d05a566/fs/statfs.c#L51

rfjakob:
Merged, thanks!

PR #1017, opened 2026-06-25
reverse: hide custom -config file located inside CIPHERDIR (#1009)

### LINKED ISSUE #1009: Reverse mode with --config does not create a plaintext config file
Hello!

I'm really starting to like gocryptfs. Thanks for making this!

One of my use cases is Reverse mode to backup my machines to a central location. The goal is to prevent a compromised backup server from being able to leak data. I recently switched from the default config file location to a custom one with with `--config`. After doing this, I noticed two things:

- `gocryptfs.conf` in the cipherdir root has gone missing
- `.gocryptfs.reverse.conf` (or whatever you named your config file with `--config`) is not hidden anymore in the cipherdir

Is this intentional? The manpage does not mention this behavior. If one does not realize this is happening, backups might  become worthless due to the missing config file (if you were naughty and didn't write down the master key).

If this is intentional behavior, I'd like to request an option that places a plaintext `gocryptfs.conf` in the cipherdir even when a custom config location is used.

Thanks a lot!

maximilize:
Fixes #1009

## Problem

In `-reverse` mode, a custom config file passed via `-config` that happens to live inside `CIPHERDIR` shows up as an ordinary **encrypted** file in the encrypted view — exactly what you flagged in the issue ("having the config file encrypted in the cipherdir seems completely useless").

By default the config (`.gocryptfs.reverse.conf`) is mapped to a virtual `gocryptfs.conf`. With `-config`, `ConfigCustom` skips that mapping (`node_dir_ops.go`, `virtualnode.go`), so an in-`CIPHERDIR` custom config is neither mapped nor hidden and leaks into the view in encrypted form.

## Repro (built from master 1a03e99)

`-config` pointing at a file **inside** the plaintext dir:

| case | encrypted view (mountpoint) |
|---|---|
| default (no `-config`) | `secret.txt`(enc), `gocryptfs.conf`, `gocryptfs.diriv` |
| `-config DIR/my.conf` (before) | **two** encrypted entries (`secret.txt` **+ the config**), `gocryptfs.diriv` |
| `-config DIR/my.conf` (after) | `secret.txt`(enc), `gocryptfs.diriv` |
| `-config OUTSIDE/my.conf` | `secret.txt`(enc), `gocryptfs.diriv` (already fine) |

## Fix

Plumb the config path into the reverse `RootNode`. When the custom config file is located inside `CIPHERDIR`, treat it as excluded in `isExcludedPlain`, so it is hidden from both readdir and lookup — the same `EPERM` path already used for `-exclude`d files. Config files outside `CIPHERDIR` are unaffected.

## Docs

Also documents the `-config` reverse-mode behaviour in `MANPAGE.md` — the missing virtual `gocryptfs.conf` (direction (1) you preferred in the issue): the custom config / master key must be backed up separately, otherwise the encrypted data cannot be decrypted.

## Tests

New `tests/reverse/config_custom_test.go`, passing for the default, `-plaintextnames` and `-deterministic-names` modes. Existing reverse tests unaffected.

---
_AI-assisted; the repro, all code references and the tests were verified manually against master._

### DIFF
--- a/Documentation/MANPAGE.md
+++ b/Documentation/MANPAGE.md
@@ -457,6 +457,14 @@ Use specified config file instead of `CIPHERDIR/gocryptfs.conf`.
 
 Applies to: all actions that use a config file: mount, `-fsck`, `-passwd`, `-info`, `-init`.
 
+In `-reverse` mode, this also changes what the encrypted view contains: by
+default the config is exposed there as a virtual `gocryptfs.conf` (so a backup
+of the encrypted view is self-contained), but with `-config` no `gocryptfs.conf`
+is presented. Make sure to back up the config file (or the master key) from its
+custom location separately, otherwise the encrypted data cannot be decrypted.
+If the custom config file is located inside `CIPHERDIR`, it is hidden from the
+encrypted view rather than exposed in encrypted form.
+
 #### -cpuprofile string
 Write cpu profile to specified file.
 
--- a/internal/fusefrontend/args.go
+++ b/internal/fusefrontend/args.go
@@ -24,6 +24,10 @@ type Args struct {
 	// location. If it is false, reverse mode maps ".gocryptfs.reverse.conf"
 	// to "gocryptfs.conf" in the plaintext dir.
 	ConfigCustom bool
+	// Config is the path to the config file. In reverse mode, a custom config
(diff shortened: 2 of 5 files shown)


rfjakob:
Merged, thanks!

Issue #1026, opened 2026-08-15
Directory listings are much slower than the backing store


redneb:
## Summary

I have a gocryptfs volume containing hundreds of thousands of files where a names-only `find` (e.g. `find -iname '*something*'`) through the mounted filesystem takes about 30 times as long as the equivalent traversal of the backing directory. This was surprising to me, because I did not expect that gocryptfs would add such a significant overheard compared to the backing store.

To isolate the problem, I created a synthetic directory containing one million short, empty files on NFS. Listing the backing directory took about 0.2 seconds, while listing the same files through gocryptfs took 25.9 seconds.

Source investigation showed that go-fuse negotiates FUSE `READDIRPLUS` by default. Unlike ordinary `READDIR`, `READDIRPLUS` also requests attributes for every entry. This makes a names-only listing perform per-entry metadata work that the application did not request.

As an A/B test, I built gocryptfs with `fuse.MountOptions.DisableReadDirPlus = true`. The mounted listing fell from 25.9 seconds to about 0.75 seconds. Re-enabling `READDIRPLUS` increased it to 13–24 seconds again, confirming it as the cause of the large difference.

## Cause

For each entry returned by gocryptfs `File.Readdirent`, go-fuse's `READDIRPLUS` path requests a lookup. Because gocryptfs does not currently implement `fs.FileLookuper`, this falls back to `Node.Lookup`, which validates and re-encrypts the name, obtains a backing directory fd, calls `fstatat`, and creates a go-fuse inode.

The result is one eager lookup and backing stat per entry, even for applications that only need names. The attributes also make each FUSE response record substantially larger and force go-fuse to create an inode for every listed entry, increasing request count and memory use.

Depending on backing metadata-cache state, a listing can be CPU-bound, storage-bound, or a mixture of both. go-fuse v2.9 defines `CAP_READDIRPLUS_AUTO`, but does not negotiate it.

## Key measurements

The following measurements used a fresh read-only gocryptfs process. Values are medians except for the single-run NFS metadata-heavy result.

| Backing | Files | Workload | Runs | `READDIRPLUS` | Disabled | Change |
| --- | ---: | --- | ---: | ---: | ---: | --- |
| NFS | 1,000,000 | Names only (`ls -U -1A`) | 3 | 20.21 s | 1.11 s | 18.2× faster |
| Local | 100,000 | Names only (`ls -U -1A`) | 5 | 0.78 s | 0.12 s | 6.5× faster |
[...]

PR #1027, opened 2026-08-15
fusefrontend: reuse readdir names for READDIRPLUS

redneb:
## Summary

Implement `fs.FileLookuper` on forward-mode directory handles so `READDIRPLUS` can reuse the name translation and backing directory file descriptor already available in `Readdirent`.

The common path now avoids repeating the directory-cache lookup, fd duplication, plaintext-name validation, and EME encryption for every listed entry. It still performs the backing `fstatat`, creates the go-fuse inode, and returns the same attributes as `Node.Lookup`.

Refs #1026.

## Correctness

The directory handle retains only the most recently returned plaintext-to-raw-ciphertext mapping:

- The mapping survives across requests for go-fuse's overflow-entry handling.
- A seek clears it.
- Interrupted-read replay or any name mismatch falls back to `Node.Lookup`.
- Long names retain the hashed name that actually exists on disk, not the full ciphertext stored in the `.name` sidecar.
- Noncanonical long-name representations fall back to `Node.Lookup` and are reported as mitigated corruption.
- `-badname` always uses the existing `Node.Lookup` path because its reverse mapping can be ambiguous.
- `ForceOwner`, `-sharedstorage` child reuse, symlink size translation, and idle-monitor activity are preserved.

## Benchmarks

Each measurement used a fresh read-only gocryptfs process and `ls -U -1A`. Variant order was alternated. The baseline was commit `790362e`; the test variant was this branch.

On an NFS-backed directory containing 1,000,000 short, empty files, across five runs:

- Median wall time: 23.86 s → 18.83 s (21% lower).
- Median gocryptfs CPU time: 11.58 s → 9.11 s (21% lower).
- Wall-time ranges were 20.43–32.10 s and 6.90–30.99 s respectively. The wide range reflects changes in NFS metadata-cache state.

On a local directory containing 100,000 short, empty files, across five runs:

- Median wall time: 0.78 s → 0.66 s (15% lower).
- Median gocryptfs CPU time: 0.80 s → 0.63 s (21% lower).

This optimization does not remove the per-entry `fstatat`, inode creation, memory use, or larger `READDIRPLUS` response records. Disabling `READDIRPLUS` remains a separate proposal.

## Test plan

- [x] `go test ./...`
- [x] `go test -race -count=1 ./internal/fusefrontend`
- [x] `go test -count=20 ./internal/fusefrontend -run 'TestFileLookup'`
- [x] `go vet ./...`
- [x] Fresh-mount NFS benchmark against the pre-change commit
- [x] Fresh-mount local-storage benchmark against the pre-change commit

### DIFF
--- a/internal/fusefrontend/file_dir_ops.go
+++ b/internal/fusefrontend/file_dir_ops.go
@@ -65,6 +65,7 @@ func (n *Node) OpendirHandle(ctx context.Context, flags uint32) (fh fs.FileHandl
 		ds:        ds,
 		dirIV:     dirIV,
 		isRootDir: n.IsRoot(),
+		node:      n,
 	}
 
 	return file, fuseFlags, errno
@@ -91,6 +92,18 @@ type DirHandle struct {
 
 	// fs.loopbackDirStream with a private dup of the file descriptor
 	ds fs.FileHandle
+
+	// lastName and lastCName are the plaintext and raw ciphertext names from
+	// the last entry returned by Readdirent. go-fuse's READDIRPLUS bridge
+	// serializes Readdirent, Lookup, and Seekdir on each file handle.
+	//
+	// Keep this mapping across calls: go-fuse may carry an overflow entry into
+	// the next READDIRPLUS request without calling Readdirent again.
+	lastName  string
+	lastCName string
+
+	// Node this directory handle was opened from.
+	node *Node
 }
 
 var _ = (fs.FileReleasedirer)((*File)(nil))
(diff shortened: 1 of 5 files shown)


rfjakob:
Looking at https://pkg.go.dev/github.com/hanwen/go-fuse/v2/fs#FileLookuper and also verified in code, fs.FileLookuper is only used for READDIRPLUS. As we now have disabled READDIRPLUS per default, I think the extra complexity here is not worth it.

redneb:
That makes sense now that `READDIRPLUS` is opt-in. The optimization may be worth reconsidering if `READDIRPLUS` or `READDIRPLUS_AUTO` becomes more prominent in the future, but I agree that the additional complexity is difficult to justify with the new default. Thanks for reviewing it and for merging the default-`READDIR` change.

PR #1028, opened 2026-08-16
Use READDIR by default and add -readdirplus

redneb:
## Summary

- Disable FUSE `READDIRPLUS` negotiation by default.
- Add `-readdirplus` to opt back into `READDIRPLUS`.
- Apply the same default to forward and reverse mounts.
- Document the performance trade-off and exercise both directory-reading modes in the test matrix.

Related to #1026. This PR is independent of #1027 and does not depend on its `FileLookuper` optimization.

## Rationale

go-fuse negotiates `READDIRPLUS` by default. It asks gocryptfs to look up every directory entry while enumerating it, even when an application only requested names. For large directories this adds per-entry name translation, backing-store metadata operations, inode allocation, CPU time, and memory use.

This PR sets `fuse.MountOptions.DisableReadDirPlus` by default, restoring demand-driven behavior: applications that only enumerate names avoid eager lookups, while applications that need metadata request it separately. `-readdirplus` retains the previous behavior for metadata-heavy workloads.

The earlier version of this PR made `-sharedstorage` imply `-noreaddirplus`. That special case is no longer needed because plain `READDIR` is now the global default. An explicit `-sharedstorage -readdirplus` enables `READDIRPLUS`; whether that helps depends on the workload and backing storage because `-sharedstorage` disables kernel attribute caching.

The default also applies to reverse mode. Backup and synchronization workloads that inspect metadata for most entries may benefit from `-readdirplus`.

## Performance trade-off

The benchmarks do not show a universally better fixed setting. Plain `READDIR` makes names-only listings 6.5–18.2× faster, while `READDIRPLUS` makes metadata-heavy listings 1.36–1.82× faster with the default cache timeouts.

The names-only improvement is much larger, and plain `READDIR` avoids metadata work until an application asks for it. Users whose workloads inspect metadata for most entries can restore the previous behavior with `-readdirplus`.

`READDIRPLUS_AUTO` would not fully address the original workload because it always starts each opened directory with `READDIRPLUS`. The real workload contains thousands of mostly small directories that fit in the first response.

## Benchmarks

All measurements are forward mode and used a fresh read-only gocryptfs process. Variant order was alternated. Names-only listings used `ls -U -1A`; metadata-heavy listings used `ls -lA`. The comparison is `-readdirplus` → default `READDIR`. Peak RSS was sampled from `/proc` every 50 ms. Values are medians except where noted as a single run.

Default cache timeouts:

| Backing | Files | Workload | Runs | Elapsed time | gocryptfs CPU time | Peak memory (RSS) |
| --- | ---: | --- | ---: | --- | --- | --- |
| NFS | 1,000,000 | Names only (`ls -U -1A`) | 3 | 20.21 s → 1.11 s (18.2× faster) | 10.54 s → 1.00 s (10.5× better) | 623,604 → 28,316 KiB (22.0× better) |
| Local | 100,000 | Names only (`ls -U -1A`) | 5 | 0.78 s → 0.12 s (6.5× faster) | 0.74 s → 0.10 s (7.4× better) | 69,292 → 21,260 KiB (3.3× better) |
| Local | 100,000 | Metadata-heavy (`ls -lA`) | 5 | 3.25 s → 4.42 s (1.36× slower) | 2.57 s → 3.25 s (1.26× worse) | — |
| NFS | 1,000,000 | Metadata-heavy (`ls -lA`) | 1 | 209.84 s → 381.87 s (1.82× slower) | 68.74 s → 112.78 s (1.64× worse) | — |

With `-sharedstorage`, the comparison is `-sharedstorage -readdirplus` → default `-sharedstorage`:

| Backing | Files | Workload | Runs | Elapsed time | gocryptfs CPU time | Peak memory (RSS) |
| --- | ---: | --- | ---: | --- | --- | --- |
| Local | 100,000 | Names only (`ls -U -1A`) | 5 | 1.00 s → 0.19 s (5.3× faster) | 1.01 s → 0.18 s (5.6× better) | 75,696 → 21,412 KiB (3.5× better) |
| Local | 100,000 | Metadata-heavy (`ls -lA`) | 5 | 18.46 s → 16.44 s (1.12× faster) | 14.20 s → 12.35 s (1.15× better) | 96,784 → 89,416 KiB (1.08× better) |
| NFS | 1,000,000 | Metadata-heavy (`ls -lA`) | 1 | 560.21 s → 629.56 s (1.12× slower) | 306.89 s → 278.10 s (1.10× better) | 537,848 → 763,736 KiB (1.42× worse) |

The NFS metadata-heavy figures are single runs, so they should not be treated as stable point estimates. They show the potential trade-off: plain `READDIR` can lose NFS bulk-attribute locality and make metadata-heavy shared-storage listings slower. The 1.42× worse peak-memory result on that single NFS run is unexplained.

## Test plan

- `go test -count=1 ./...`
- `go vet ./...`
- Added CLI parsing coverage for both `-readdirplus` and `-o readdirplus`.
- Added mount-option coverage for the default, explicit `-readdirplus`, and `-sharedstorage -readdirplus`.
- The normal test matrix now exercises plain `READDIR`; an explicit `-readdirplus` variant preserves `READDIRPLUS` coverage.
- Manually verified on Linux 6.18 with `-fusedebug` that the FUSE INIT reply omits `READDIRPLUS` by default and advertises it with `-readdirplus`, including in combination with `-sharedstorage`.

### DIFF
--- a/Documentation/MANPAGE.md
+++ b/Documentation/MANPAGE.md
@@ -378,6 +378,9 @@ metadata-heavy workloads such as `ls -l`. With the default cache
 timeouts (that is, without `-sharedstorage`), `ls -l` was 36% slower
 on local storage and 82% slower in a single NFS run.
 
+`-sharedstorage` implies `-noreaddirplus` and this cannot be
+overridden.
+
 This option currently only has an effect on Linux.
 
 For benchmarks and more details, see
@@ -434,7 +437,7 @@ Enable work-arounds so gocryptfs works better when the backing
 storage directory is concurrently accessed by multiple gocryptfs
 instances.
 
-At the moment, it does two things:
+At the moment, it does three things:
 
 1. Disable stat() caching so changes to the backing storage show up
    immediately.
@@ -442,6 +445,11 @@ At the moment, it does two things:
    storage are not stable when files are deleted and re-created behind
    our back. This would otherwise produce strange "file does not exist"
    and other errors.
+3. Disable FUSE `READDIRPLUS`, because its eagerly fetched attributes
+   cannot be reused by the kernel's attribute cache when stat caching
+   is disabled. The backing filesystem may still cache attributes.
+   On network backing storage, metadata-heavy directory listings may
+   become slower.
 
 When "-sharedstorage" is active, performance is reduced and hard
 links cannot be created.
--- a/Documentation/MANPAGE.md
+++ b/Documentation/MANPAGE.md
@@ -366,6 +366,23 @@ out-of-space errors for a massive speedup.
 For benchmarks and more details of the issue see
 https://github.com/rfjakob/gocryptfs/issues/63 .
 
+#### -noreaddirplus
+Disable FUSE `READDIRPLUS`. By default, Linux may request attributes
+for every entry during a directory listing. This improves workloads
+that immediately inspect most entries, but adds unnecessary work to
+names-only listings.
+
+This option restores demand-driven behavior: attributes are fetched
+only when the application requests them. It can substantially slow
+metadata-heavy workloads such as `ls -l`. With the default cache
+timeouts (that is, without `-sharedstorage`), `ls -l` was 36% slower
+on local storage and 82% slower in a single NFS run.
+
+This option currently only has an effect on Linux.
+
+For benchmarks and more details, see
+https://github.com/rfjakob/gocryptfs/issues/1026 .
+
 #### -nosuid
 See `-suid, -nosuid`.
 
--- a/Documentation/MANPAGE.md
+++ b/Documentation/MANPAGE.md
@@ -366,26 +366,6 @@ out-of-space errors for a massive speedup.
 For benchmarks and more details of the issue see
 https://github.com/rfjakob/gocryptfs/issues/63 .
 
-#### -noreaddirplus
-Disable FUSE `READDIRPLUS`. By default, Linux may request attributes
-for every entry during a directory listing. This improves workloads
-that immediately inspect most entries, but adds unnecessary work to
-names-only listings.
-
-This option restores demand-driven behavior: attributes are fetched
-only when the application requests them. It can substantially slow
-metadata-heavy workloads such as `ls -l`. With the default cache
-timeouts (that is, without `-sharedstorage`), `ls -l` was 36% slower
-on local storage and 82% slower in a single NFS run.
-
-`-sharedstorage` implies `-noreaddirplus` and this cannot be
-overridden.
-
-This option currently only has an effect on Linux.
-
-For benchmarks and more details, see
-https://github.com/rfjakob/gocryptfs/issues/1026 .
-
 #### -nosuid
 See `-suid, -nosuid`.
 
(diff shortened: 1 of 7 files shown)


rfjakob:
Looks like readdirplus_auto may the silver bullet here. Juicefs, which uses a patched go-fuse supports it ( https://juicefs.com/docs/community/fuse_mount_options/#readdirplusauto ). Would be worth looking at the patch that adds support, and upstreaming it.

redneb:
I agree that `READDIRPLUS_AUTO` would generally be a better default for gocryptfs than today’s always-on `READDIRPLUS`. It retains eager attributes when applications consume them, but can switch to plain `READDIR` when they do not. I plan to attempt a go-fuse PR adding support for it.

However, AUTO is not a complete solution for names-only traversal. It operates independently per opened directory and always uses `READDIRPLUS` for the first request. It can switch to plain `READDIR` only if the directory needs another request and no lookup occurred in between. A directory that fits in the first response therefore receives no benefit: every entry still incurs gocryptfs’s eager lookup, backing `fstatat`, and inode creation.

This limitation applies directly to my original workload. The real volume contains 212,565 entries spread across 6,427 mostly small directories: the median contains 17 entries and the largest contains 3,374. With the mount’s 1 MiB NFS buffer, even the largest directory’s complete `READDIRPLUS` payload is at most about 600 KiB. Every directory therefore fits in its mandatory first PLUS response, so AUTO would avoid essentially none of the eager lookups during my names-only `find`. If a directory did require multiple requests, the subdirectory lookups needed for recursive traversal could also cause AUTO to select `READDIRPLUS` again.

I therefore see AUTO and `-noreaddirplus` as complementary. AUTO would be the better adaptive default, while `-noreaddirplus` provides a deterministic escape hatch for workloads involving many small directories where the heuristic cannot switch soon enough. Since the flag remains opt-in, accepting it now would not prevent AUTO from becoming the default later.

rfjakob:
I looked at the kernel code, matches what you said.

But looking at https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/fs/fuse/dir.c?id=4582a4ab2a0e7218449fb2e895d0aae9ea753c94 there's another option. Quoting the commit msg: 

> If the user space implementation wants to disable readdirplus
temporarily, it could just return ENOTSUPP. Then kernel will
recall it with readdir.

gocryptfs could make readdirplus adaptive this way, biasing towards readdir.

redneb:
This looked promising, so I tested it on Linux 6.18.35 with a minimal raw go-fuse filesystem that supports `READDIR` but returns `EOPNOTSUPP` for `READDIRPLUS`. The directory read failed with `EOPNOTSUPP`; the trace showed the rejected `READDIRPLUS` but no subsequent `READDIR`.

This also matches [`fuse_readdir_uncached()`](https://github.com/gregkh/linux/blob/v6.18.35/fs/fuse/readdir.c#L338-L385): it selects one opcode, calls `fuse_simple_request()`, and returns a negative result unchanged. I cannot find an `EOPNOTSUPP` retry in that function, elsewhere in `fs/fuse/readdir.c`, or in the VFS `iterate_dir()` path. The cited commit appears to add the AUTO selection heuristic, but not the fallback described in its commit message.

The test did not negotiate `READDIRPLUS_AUTO`, because current go-fuse does not support that. However, even with AUTO, offset zero always selects `READDIRPLUS`, and I cannot see any code that retries a failed request. A later caller-initiated read at a nonzero offset might select `READDIR`, but the current `getdents64()` call has already failed and normal callers do not retry it.

Am I missing another retry path? Otherwise, returning `EOPNOTSUPP` does not appear to provide a usable adaptive fallback on current kernels, so it would not replace `-noreaddirplus`.

rfjakob:
Too bad.

In that case, please disable readdirplus per default, and add a "-readdirplus" flag to re-enable.

redneb:
Implemented and pushed the requested change in d9d9773:

- Plain `READDIR` is now the default.
- `-readdirplus` restores the previous `READDIRPLUS` behavior.
- The default applies to both forward and reverse modes.
- An explicit `-sharedstorage -readdirplus` combination is supported.
- Documentation and tests have been updated; the full test suite and `go vet ./...` pass.

The branch now has three commits because the first two represent the earlier design. Would you prefer that I rewrite the branch into one clean commit, or leave the history as-is?

lechner:
Hi, thanks for this!  With these commits I can finally use my encrypted Maildir folder on NFS!

rfjakob:
Merged as https://github.com/rfjakob/gocryptfs/commit/5b2881e7cbc347eb379d35c3ae1f491bef8f227f , thanks!

PR #1029, opened 2026-08-17
fusefrontend: take ContentLock for truncate through node.Setattr

### LINKED ISSUE #1024: Truncating file after writing at least 4096 bytes results in corrupt block and I/O error
# Detailed description

Hello, there seems to be an issue with gocryptfs:

1. Create and open an empty file.
2. Write at least 4096 bytes (block size on ext4, may be different?).
3. Truncate the file (shrink).
4. Write some data of any length.
5. Try truncating (shrinking) file again.

Last step fails with `Input/output error` errno, other write operations (e.g. `write` or `ftruncate`) fail too.

Journal log says the operation failed due to corrupted block:

```log
Aug 11 11:33:06 ubuntu gocryptfs[1404]: doRead 543573: corrupt block #0: cipher: message authentication failed
Aug 11 11:33:06 ubuntu gocryptfs[1404]: Truncate: shrink doRead returned error: <nil>
```

# System info

Tested `gocryptfs` versions: `2.6.1`, `master` (commit `790362e8f8eabd1bdf75f98328daf94cf623c600`)

```json
{
	"Hostname" : "ubuntu",
	"StaticHostname" : "ubuntu",
	"PrettyHostname" : null,
	"DefaultHostname" : "localhost",
	"HostnameSource" : "static",
	"IconName" : "computer-vm",
	"Chassis" : "vm",
	"Deployment" : null,
	"Location" : null,
	"KernelName" : "Linux",
	"KernelRelease" : "6.17.0-29-generic",
	"KernelVersion" : "#29-Ubuntu SMP PREEMPT_DYNAMIC Tue May  5 19:42:34 UTC 2026",
	"OperatingSystemPrettyName" : "Ubuntu 25.10",
	"OperatingSystemCPEName" : null,
	"OperatingSystemHomeURL" : "https://www.ubuntu.com/",
	"OperatingSystemReleaseData" : [
		"PRETTY_NAME=Ubuntu 25.10",
		"NAME=Ubuntu",
		"VERSION_ID=25.10",
		"VERSION=25.10 (Questing Quokka)",
		"VERSION_CODENAME=questing",
		"ID=ubuntu",
		"ID_LIKE=debian",
		"HOME_URL=https://www.ubuntu.com/",
		"SUPPORT_URL=https://help.ubuntu.com/",
		"BUG_REPORT_URL=https://bugs.launchpad.net/ubuntu/",
		"PRIVACY_POLICY_URL=https://www.ubuntu.com/legal/terms-and-policies/privacy-policy",
		"UBUNTU_CODENAME=questing",
		"LOGO=ubuntu-logo"
	],
}
```

Also tested on Ubuntu 24.04.3 LTS.

# How to reproduce

```c
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <unistd.h>

#define BUFFER_LEN 4096 // should be at least 4096 (block size)

int main() {
  int status;
  int fd;

  int buffer[BUFFER_LEN + 1] = {};
  for (int i = 0; i < BUFFER_LEN; ++i) {
    buffer[i] = i;
  }

  status = creat("file", S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
  printf("CREAT: %d\n", status);
  close(status);

  status = open("file", O_WRONLY);
  printf("OPEN: %d\n", status);
  fd = status;

  // bug only appears if writing >= 4096 bytes 
  status = write(fd, buffer, BUFFER_LEN);
  printf("WRITE: %d\n", status);

  // size arguments here and below are not specific
  status = truncate("file", 8);
  printf("TRUNCATE: %d\n", status);

  status = write(fd, "Test data!", 10);
  printf("WRITE: %d\n", status);

  status = truncate("file", 7);
  printf("TRUNCATE: %d (%s)\n", status, strerror(errno));
}

// ::Expected::
// CREAT: 3
// OPEN: 3
// WRITE: 4096
// TRUNCATE: 0
// WRITE: 10
// TRUNCATE: 0 (Success)

// ::Actual::
// CREAT: 3
// OPEN: 3
// WRITE: 4096
// TRUNCATE: 0
// WRITE: 10
// TRUNCATE: -1 (Input/output error)
```

maximilize:
Fixes #1024.

`truncate(2)` reaches `node.Setattr`, which opens its own file handle and calls `f2.truncate(sz)` directly (`internal/fusefrontend/node.go:249`). `file.Setattr` does the same work while holding `ContentLock` (`file_setattr.go:29`), and `Allocate` holds it too (`file_allocate_truncate.go:58`) — the node path is the only caller that does not, although `truncate` ends in `doWrite`, whose comment states the caller has locked it.

Two things follow from the missing lock, and the second is what the issue reports:

1. A path truncate is not serialized against concurrent writes to the same file.
2. `ContentLock.Lock()` is what increments the global write-operation counter (`openfiletable.countingMutex.Lock`). `isConsecutiveWrite` uses that counter to notice that somebody else touched the file since this handle's last write. A truncate through the node path leaves the counter untouched, so an already-open handle still thinks its next write appends, skips `writePadHole`, and grows the file past a short last block. The block keeps the authentication tag of its old length and no longer decrypts:

```
doRead 543573: corrupt block #0: cipher: message authentication failed
```

That is the reporter's sequence exactly: write 4096 bytes, `truncate(file, 8)` (the descriptor keeps its offset of 4096), write 10 bytes — which now lands past the new EOF — and every later read, write or truncate of that file fails with EIO.

The fix adds `truncateLocked`, which takes `fdLock`/`ContentLock` and then calls `truncate`, and has `node.Setattr` use it. `file.Setattr` and `Allocate` keep their existing locking, so nothing takes the lock twice.

Verified without a mount: `internal/fusefrontend/issue_1024_test.go` drives the File methods against a backing file and repeats the reporter's sequence through a second handle, the way `node.Setattr` does. Against the unchanged code the test fails with `corrupt block #0` and EIO; with the fix the first block reads back as the expected 8 bytes plus zero padding, the payload at 4096 is intact, and shrinking again succeeds. `go test -tags without_openssl ./internal/...` passes for 13 packages with and without the change; the one failure in both is `TestGetdents`, which is unrelated (it inspects `/dev` inside the container).

### DIFF
--- a/internal/fusefrontend/file_allocate_truncate.go
+++ b/internal/fusefrontend/file_allocate_truncate.go
@@ -96,7 +96,27 @@ func (f *File) Allocate(ctx context.Context, off uint64, sz uint64, mode uint32)
 	return f.truncateGrowFile(oldPlainSz, newPlainSz)
 }
 
+// truncateLocked takes the locks that truncate needs and truncates.
+//
+// truncate calls doWrite, which requires ContentLock to be held by the caller.
+// file.Setattr already holds it; node.Setattr (the path used by truncate(2))
+// opens a fresh file handle and has nothing around it, so it goes through here.
+func (f *File) truncateLocked(newSize uint64) syscall.Errno {
+	f.fdLock.RLock()
+	defer f.fdLock.RUnlock()
+	if f.released {
+		tlog.Warn.Printf("ino%d fh%d: truncate on released file", f.qIno.Ino, f.intFd())
+		return syscall.EBADF
+	}
+	f.fileTableEntry.ContentLock.Lock()
+	defer f.fileTableEntry.ContentLock.Unlock()
+
+	return f.truncate(newSize)
+}
+
 // truncate - called from node.Setattr and file.Setattr.
+//
+// The caller must hold ContentLock, see truncateLocked.
 func (f *File) truncate(newSize uint64) (errno syscall.Errno) {
 	var err error
 	// Common case first: Truncate to zero
--- a/internal/fusefrontend/issue_1024_test.go
+++ b/internal/fusefrontend/issue_1024_test.go
@@ -1,0 +1,116 @@
+package fusefrontend
+
+// Regression test for https://github.com/rfjakob/gocryptfs/issues/1024
+//
+// truncate(2) goes through node.Setattr, which opens its own file handle. That
+// handle must take ContentLock like file.Setattr does: the lock doubles as the
+// global write-operation counter that isConsecutiveWrite() uses to notice that
+// somebody else changed the file. Without it, an already-open handle keeps
+// believing its next write appends, skips writePadHole(), and leaves the last
+// block short while the file grows past it - the block then fails to decrypt.
+
+import (
+	"bytes"
+	"context"
+	"path/filepath"
+	"syscall"
+	"testing"
+
+	"github.com/rfjakob/gocryptfs/v2/internal/contentenc"
+	"github.com/rfjakob/gocryptfs/v2/internal/cryptocore"
+	"github.com/rfjakob/gocryptfs/v2/internal/nametransform"
+)
+
+// issue1024Fs sets up a fusefrontend root over a temporary directory.
+func issue1024Fs(t *testing.T) (*RootNode, string) {
+	t.Helper()
+
+	cipherdir := t.TempDir()
+	key := make([]byte, cryptocore.KeyLen)
+	cCore := cryptocore.New(key, cryptocore.BackendGoGCM, contentenc.DefaultIVBits, true)
+	cEnc := contentenc.New(cCore, contentenc.DefaultBS)
+	nt := nametransform.New(cCore.EMECipher, true, 255, true, nil, false)
+
+	return NewRootNode(Args{Cipherdir: cipherdir}, cEnc, nt), filepath.Join(cipherdir, "backing")
+}
+
+// issue1024Open opens path as a File, like node.Open would.
+func issue1024Open(t *testing.T, rn *RootNode, path string, flags int) *File {
+	t.Helper()
+
+	fd, err := syscall.Open(path, flags, 0600)
+	if err != nil {
+		t.Fatalf("open %q: %v", path, err)
+	}
+	f, _, errno := NewFile(fd, path, rn)
+	if errno != 0 {
+		t.Fatalf("NewFile: %v", errno)
+	}
+	t.Cleanup(func() {
+		if !f.released {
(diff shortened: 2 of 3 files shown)


rfjakob:
The test passes even when I revert the change

```
0 jakob@brikett:/ssd2/jakob.donotbackup/code/gocryptfs/internal/fusefrontend$ git diff
diff --git a/internal/fusefrontend/node.go b/internal/fusefrontend/node.go
index b1c25be..28ebbd5 100644
--- a/internal/fusefrontend/node.go
+++ b/internal/fusefrontend/node.go
@@ -246,7 +246,7 @@ func (n *Node) Setattr(ctx context.Context, f fs.FileHandle, in *fuse.SetAttrIn,
                }
                f2 := f.(*File)
                defer f2.Release(ctx)
-               errno = syscall.Errno(f2.truncateLocked(sz))
+               errno = syscall.Errno(f2.truncate(sz))
                if errno != 0 {
                        return errno
                }

0 jakob@brikett:/ssd2/jakob.donotbackup/code/gocryptfs/internal/fusefrontend$ go test -v
test_helpers: warning: testParentDir "/tmp/gocryptfs-test-parent-1026" does not reside on ext4, we will miss failures caused by ino reuse
=== RUN   TestTruncateThroughSecondHandleKeepsBlocksValid
--- PASS: TestTruncateThroughSecondHandleKeepsBlocksValid (0.00s)
=== RUN   TestPrepareAtSyscall
    prepare_syscall_test.go:19: cipherdir = "/tmp/gocryptfs-test-parent-1026/2764118719/TestPrepareAtSyscall.2945655390"
--- PASS: TestPrepareAtSyscall (0.01s)
=== RUN   TestPrepareAtSyscallPlaintextnames
--- PASS: TestPrepareAtSyscallPlaintextnames (0.01s)
=== RUN   TestEncryptDecryptXattrName
Could not stat backing directory "": no such file or directory
DetectQuirks: Statfs on "" failed: no such file or directory
    xattr_unit_test.go:40: cAttr=user.gocryptfs.22EILvJ6YPKH9RSSHlPZR3aG0I4fyay43G3MUVQNi6c
--- PASS: TestEncryptDecryptXattrName (0.00s)
PASS
ok  	github.com/rfjakob/gocryptfs/v2/internal/fusefrontend	0.026s

```

rfjakob:
Somewhat edited and merged as 9c2130d6c2e54c45c6787fbc49a896a8544a5e47, thanks!

